PR #15, opened 2015-10-14
made my own directory

tycobass:


PythonCHB:
working too fast during class -- didn't notice that you created a dir called "Initial_Python_directory", whic is OK in itself, but you put it in the students dir -- so a bit out of place.

try renaming that to your name (or some name :-) ) and submitting a new pull request. Somethign like:

git mv Initial_Python_directory the_name_I_want_to_use

git commit -a -m "renaming my dir"

git push 

[go to github to do the pull request]

PR #19, opened 2015-10-14
added bob

HasBob:


students/bob/box.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/python
```
PythonCHB: is this running in Python3?

the input() function is a bit different -- in python2, it will evaluate what is typed so that "3" gets evaluated to an integer.

in py3, it simply gives the the actual text typed, so you have to do: int( input ('a msg>>')) if you want an integer.

PR #21, opened 2015-10-14
added grid_printer.py

katdev:


PythonCHB:
I needed to do a bit of cleanup -- you had a kathleen dir both in teh main dir, and under students.

should be all good now. Do a:

git pull upstream master
git commit -a
git push origin master

and it should all be in sync.

PR #26, opened 2015-10-14
grid file yey

austinscara:
the grid

### DIFF
--- a/students/AustinScara/FizzBuzz.py
+++ b/students/AustinScara/FizzBuzz.py
@@ -1,0 +1,18 @@
+def divThree(n):
+	return n % 3 == 0
+
+
+def divFive(n):
+	return n % 5 == 0
+
+
+
+for number in range(101):
+	if divThree(number) is True and divFive(number) is True:
+		print ("FizzBuzz")
+	elif divThree(number) is True and divFive(number) is False:
+		print ("Fizz")
+	elif divThree(number) is False and divFive(number) is True:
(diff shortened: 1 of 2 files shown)


students/AustinScara/FizzBuzz.py
```diff
@@ -0,0 +1,17 @@
+def divThree(n):
```
PythonCHB: nice breaking it down.

students/AustinScara/FizzBuzz.py
```diff
@@ -0,0 +1,17 @@
...
+	return n % 3 == 0
+
+
+def divFive(n):
+	return n % 5 == 0
+
+
+for number in range(101):
+	if divThree(number) and divFive(number):
```
PythonCHB: see if you can do it without calling divThree and divFive multiple times.

students/AustinScara/grid_printer.py
```diff
@@ -0,0 +1,10 @@
+def grid(n):
+	headMidBot = "+ " + "- "*n + "+" + " -"* n+ " +"
+	mid = "|" + "  "*n + " + " + "  " *n + "|" + '\n'
+	return headMidBot +'\n' + mid*n + headMidBot + '\n' + mid*n + headMidBot
+
+
+
+size = int(input("How big Do you want this grid to be?"))
```
PythonCHB: nice!

PR #29, opened 2015-10-14
FizzBuzz program

gabrielx52:


students/gabriel_meringolo/session_02/FizzBuzz.py
```diff
@@ -0,0 +1,9 @@
+for i in range(1,101):
+    if i % 3 == 0 and i % 5 == 0:
+        print("FizzBuzz")
+    elif i % 3 == 0:
+        print("Fizz")
+    elif i % 5 == 0:
+        print("Buzz")
+    else:
+        print(i)
```
PythonCHB: this is a fine, if basic solution. But if you want to take it one step further, see if you can avoid checking for multiples of three and 5 twice.

students/gabriel_meringolo/session_02/series.py
```diff
@@ -0,0 +1,41 @@
+def fibonacci(n):
+    '''returns the nth number in the fibonacci sequence'''
+    fib = (0,1) # sets up fib sequence
+    assert type(n) == int and n > 0, "ERROR: Argument must be IntType greater than 0."
```
PythonCHB: for the most part, assert() is for testing, not type or value checking in a function call.

The "right" way to do this is to raise an Exception if n <0 -- but we haven't learned that yet, so for now, return None if n < 0

students/gabriel_meringolo/session_02/series.py
```diff
@@ -0,0 +1,41 @@
...
+    return (luc[n - 1]) # returns the nth number in the sequence, misusing one because seq starts at 0
+
+
+
+def sum_series(n, x = 0, y = 1):
+    '''returns nth number of fib sequence if x, y are default/ lucas numbers if x = 2, y = 1'''
+    if x == 0 and y == 1:
+        return fibonacci(n) # calls fib function if x, y are default
+    elif x == 2 and y == 1:
+        return lucas(n) # calls lucas function if x = 2 and y = 1
```
PythonCHB: this will work, but not that general -- try making it use the x and y values to start the series, rather than hard-coding those particular cases. In fact, you can re-write lucas and fibanacci to call sum_series -- you want to go from the specific to the general, not the other way around...

students/gabriel_meringolo/session_02/series.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def lucas(n):
+    '''returns the nth number in the lucas sequence'''
+    luc = (2,1) # sets up lucas sequence
+    assert type(n) == int and n > 0, "ERROR: Argument must be IntType greater than 0."
+    # assert checks for Int greater than zero, if false throws error
+    for i in range(n):
+        newLuc = (luc[-2] + luc[-1]) # adds the last two numbers in sequence, creates next number...
+        luc += (newLuc,) # creates new tuple with old one and new number
```
PythonCHB: we'll get into lists more next week, but here you realy want to use alist, rather than a tuple --:

then it would be: luc.append(newLuc)

PythonCHB:
Sorry I wasn't clear about where to put the asserts. The idea was to write some test code at the bottom of your module , something like:

assert fib(0) == 0
assert fib(1) == 1
assert fib(2) == 3
...

I had hoped to demo this in class, but all that darn git wrangling too a lot of time :-(

students/gabriel_meringolo/session_02/FizzBuzz.py
gabrielx52: I attempted to remove the redundancies in this for an hour or so and couldn't figure it out so I asked google... What I found ended up being a mathematical solution '''not my strong point''' - if i % 15 == 0: print("FizzBuzz").

I also came across a one liner that I don't fully understand, might be the math thing again. I don't expect a full explanation I'm sure I'll understand it in time I'm just wondering if the * is multiplication or is it being used in a different manor like an overloaded operator or something like that..
for i in range(1,101): print("Fizz"*(i%3==0) + "Buzz"*(i%5==0) or i)

Thanks
GM

PythonCHB: yup -- that's the "factor" math -- if a number is divisible by both 3 and 5, it must be divisible by 3*5, i.e. 15. But you can remove that check as well -- I'm surprised Google didn't find that for you. I'll post my solutions in a day or too, but look at that one-liner for a hint -- you can do something similar with if..elif..else.

As for teh one-liner, I think there's a typo in it -- I dont see the * -- but I think it shoudl be:
```
for i in range(1,101): print("Fizz"*(i%3==0) + "Buzz"*(i%5==0) or i)
```
in that case, yes, * is regular multiplication, but the trick here is that True and False are actually sub-types of integers (1 and 0):
```
In [212]: True * 3
Out[212]: 3

In [213]: False*3
Out[213]: 0
```
and:
```
In [215]: i = 6

In [216]: (i%3==0)
Out[216]: True

In [217]: i = 5

In [218]: (i%3==0)
Out[218]: False
```
Do you see what's going on now?

gabrielx52: Ahhhhh, I get it now. Its boolean.. I could't see that before. Thanks! 
 
I tried string concatenation for the fizz+buzz I just count seem to get it work, I'm sure its staring me in the face. I think I was putting it in the wrong place in a loop.. Thanks for the insight and help!

PR #30, opened 2015-10-14
grid printer exercise

robalford:


students/robalford/session02/fizz_buzz.py
```diff
@@ -0,0 +1,10 @@
+def fizz_buzz():
+    for i in range(1,101):
+        if i % 3 == 0 and i % 5 == 0:
```
PythonCHB: does the job -- but see if you can do it without checking for i%3 and i%5 twice...

students/robalford/session02/grid_printer.py
```diff
@@ -0,0 +1,13 @@
+def print_grid(size, rows_and_columns):
```
PythonCHB: looks good.

PR #32, opened 2015-10-15
wrote series.py and refactored fizz_buzz.py

robalford:


students/robalford/session02/fizz_buzz.py
```diff
@@ -1,10 +1,12 @@
 def fizz_buzz():
     for i in range(1,101):
-        if i % 3 == 0 and i % 5 == 0:
-            print('FizzBuzz')
-        elif i % 3 == 0:
-            print('Fizz')
-        elif i % 5 == 0:
-            print('Buzz')
+        fizz = ''
```
PythonCHB: this could be a tiny bit more compact, but that's the idea -- nice work.

students/robalford/session02/series.py
```diff
@@ -0,0 +1,27 @@
...
+def lucas(n):
+    """ Return the nth value in the Lucas Numbers Series. """
+    if n == 0:
+        return 2
+    elif n == 1:
+        return 1
+    else:
+        return lucas(n-1) + lucas(n-2)
+
+def sum_series(n, step_one=0, step_two=1):
```
PythonCHB: perfect! nice job.

PR #34, opened 2015-10-16
Added main block

gabrielx52:
I added an if statement to check if being ran directly, if so will run assert statements


PythonCHB:
very nice.

PR #35, opened 2015-10-16
fizzbuzz compacted + tests for series.py

robalford:


students/robalford/session02/fizz_buzz.py
```diff
@@ -1,12 +1,12 @@
 def fizz_buzz():
```
PythonCHB: very nice!

PythonCHB:
nicely done

PR #36, opened 2015-10-17
added fizzbuzz

HasBob:


students/bob/fizzbuzz.py
```diff
@@ -0,0 +1,27 @@
...
+## completed 2015-10-15
+
+## print 1-100 inclusive, but if divisible by 3 print "Fizz", 
+## if divisible by 5 prting "Buzz", and if divisible by both 
+## print "FizzBuzz"
+
+
+def fb():
+    x = 0
+    while x < 100:
```
PythonCHB: this really calls for a for loop:

for i in range(101):

But fine solution -- now try and see if you can not check for x % 5 == 0 twice.

PythonCHB:
Accepting pull request -- if you do decide to update it, you can do a new PR

PR #37, opened 2015-10-17
bool lab starter

gabrielx52:
I just want to make sure I'm understanding the instructions correctly before I get in too deep... 
This is what you are looking for with directly returning boolean results? 
Thanks.


PythonCHB:
Exactly.

PR #38, opened 2015-10-17
FizzBuzz Exercise - Bruce B

Baumanbg:


students/Bruce/session_02/fizzbuzz.py
```diff
@@ -0,0 +1,7 @@
+#def main():
+for n in range(100):
```
PythonCHB: note that this will go to 99 -- not 100.

Fine solution -- now see if you can avoid checking for %3 and %5 twice....

or just move on :-)

PR #39, opened 2015-10-18
series exercise done

iampei:


students/pei/session02/codingbat.py
```diff
@@ -0,0 +1,50 @@
+#warmup 1
+def sleep_in(weekday, vacation):
```
PythonCHB: nice!

students/pei/session02/codingbat.py
```diff
@@ -0,0 +1,50 @@
...
+    return (not weekday) or vacation
+
+def monkey_trouble(a_smile, b_smile):
+    return (a_smile) == (b_smile)
+# quick tests
+print (monkey_trouble(True, True))
```
PythonCHB: this would be a good palce to put an "assert":

assert monkey_trouble(True, True) is True

If that's the right answer :-)

students/pei/session02/series.py
```diff
@@ -0,0 +1,47 @@
...
+        return fib( n-2) + fib (n-1)
+
+# quick tests
+#print (fib(1))
+#print (fib(2))
+#print (fib(3))
+#print (fib(4))
+#print (fib(5))
+#print (fib(6))
+assert (fib(1)) == 1
```
PythonCHB: no need for parens around the fib(1) -- but it doesn't hurt, either.

students/pei/session02/series.py
```diff
@@ -0,0 +1,47 @@
...
+assert (fib(6)) != 5
+assert (fib(6)) == 8
+print ('Fibonacci number quick tests were successful')
+
+def lucas(n):
+    """ Lucas Numbers Exercise: starting with the value 2 and 1, instead of 0 and 1
+    """
+    if n  == 0:
+        return 2
+    if n == 1:
```
PythonCHB: I'd make this an elif -- though it doesn't make a difference in this case.

PR #40, opened 2015-10-18
Sunday pull request

JosephVC:
Adding homework files

### DIFF
--- a/students/JosephC/Session_02/FizzBuzz.py
+++ b/students/JosephC/Session_02/FizzBuzz.py
@@ -1,0 +1,29 @@
+#Intro to Python (via the University of Washington)
+
+#Fizzbuzz is said to be a problem many interviewees cannot adequately answer
+#Odd, isn't it?
+
+#Completed by: Joseph Cardenas, 10.17.2015
+
+
+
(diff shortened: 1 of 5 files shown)


students/JosephC/Session_02/FizzBuzz.py
```diff
@@ -0,0 +1,29 @@
...
+#Odd, isn't it?
+
+#Completed by: Joseph Cardenas, 10.17.2015
+
+
+
+def fizzbuzz():
+    n = 0 
+       
+    while n != 100:
```
PythonCHB: maybe use a for loop here.

students/JosephC/Session_02/FizzBuzz.py
```diff
@@ -0,0 +1,29 @@
...
+        n += 1
+        if n % 3 == 0 and n % 5 == 0:
```
PythonCHB: see if you can do this without checking for  ``n % 3 == 0 and n % 5 == 0`` twice.

PythonCHB:
IT looks like you're not done with grid printer -- so I won't merge yet.

PR #41, opened 2015-10-18
Updated grid_printer 

maxmac12:
Contains two print_grid functions that take arguments used to draw a simple grid pattern.


PythonCHB:
nice!

PR #42, opened 2015-10-18
Fizz_Buzz and Series pull

ryanmorin:
Pull request

### DIFF
--- a/students/ryan_morin/session_02/fizz_buzz.py
+++ b/students/ryan_morin/session_02/fizz_buzz.py
@@ -1,0 +1,18 @@
+__author__ = 'Ryan'
+
+def FizzBuzz(num):
+    """
+    :param num: number 100
+    :return: print values and string based on logic statements
+    """
+    for num in range(1, (num + 1)):
+        if num%3 == 0 and num%5 == 0:
+            print ("FizzBuzz")
+        elif num%3 == 0:
+            print ("Fizz")
+        elif num%5 == 0:
+            print ("Buzz")
+        else:
+            print (num)
+
+print(FizzBuzz(100))
--- a/students/ryan_morin/session_02/series.py
+++ b/students/ryan_morin/session_02/series.py
@@ -1,0 +1,40 @@
+__author__ = 'Ryan'
+
+def fibonacci(n):
+    """
+    :param n: number
+    :return: a list of nth value in the fibonacci series
+    """
+    fib_lst = [0,1]
+    for i in range(n):
+        fib_new = int(fib_lst[i]) + int(fib_lst[i+1])
+        fib_lst.append(fib_new)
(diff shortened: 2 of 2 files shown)


students/ryan_morin/session_02/fizz_buzz.py
```diff
@@ -0,0 +1,18 @@
+__author__ = 'Ryan'
+
+def FizzBuzz(num):
+    """
+    :param num: number 100
+    :return: print values and string based on logic statements
+    """
+    for num in range(1, (num + 1)):
```
PythonCHB: Solid solution. 

If you have some time, see if you can do it without repeating the num%3 and num%5.

students/ryan_morin/session_02/series.py
```diff
@@ -0,0 +1,40 @@
...
+    lucas_lst = [2,1]
+    for i in range(n):
+        lucas_new = int(lucas_lst[i]) + int(lucas_lst[i+1])
+        lucas_lst.append(lucas_new)
+    return lucas_lst[(len(lucas_lst)-2)]
+
+def sum_series(elmnt, op1=0, op2=1):
+    """
+    :param elmnt: number
+    :param op1:  optional variable sequence start value - default equal to zero
```
PythonCHB: Nice doctring!

The standard is to be a bot more compact for default values:

:param op1=0: optional variable sequence start value

students/ryan_morin/session_02/series.py
```diff
@@ -0,0 +1,40 @@
...
+def fibonacci(n):
+    """
+    :param n: number
+    :return: a list of nth value in the fibonacci series
+    """
+    fib_lst = [0,1]
+    for i in range(n):
+        fib_new = int(fib_lst[i]) + int(fib_lst[i+1])
+        fib_lst.append(fib_new)
+    return fib_lst[(len(fib_lst)-2)]
```
PythonCHB: a bit awkward to compute an extra value that you don't use.

Also -- we'll go over this this week, but you can access the second to last element with:

``fib_list[-2]``

students/ryan_morin/session_02/series.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    lst = [op1, op2]
+    for i in range(elmnt):
+        num_new = int(lst[i]) + int(lst[i+1])
+        lst.append(num_new)
+    return lst[(len(lst)-2)]
+
+assert sum_series(5)
+assert sum_series(5,2,1)
+assert sum_series(5,4,8)
```
PythonCHB: these assert that you get a non-zero (Truthy) value, but not that it's correct!

PythonCHB:
nicely done -- but do make sure you know how to do  this with an recursive, rather than iterative approach. i.e. calling the function from itself.

PR #43, opened 2015-10-19
Added fibonacci.py and fizzbuzz.py

mcimino2015:


mcimino/fibonacci.py
```diff
@@ -0,0 +1,13 @@
+# fibonacci.py
+
+
+def fibonacci(x):
```
PythonCHB: perfect.

mcimino/fizzbuzz.py
```diff
@@ -0,0 +1,12 @@
+# fizzbuzz.py
+
+
+for i in range(1, 101):
```
PythonCHB: good, solid solution -- but could you do it without checking for %3 and %5 twice?

mcimino/grid_printer.py
```diff
@@ -0,0 +1,16 @@
+def print_top(n):
```
PythonCHB: I like the breaking it out into functions.

PR #44, opened 2015-10-19
My FizzBizz solution

tycobass:
This is my first pull request.  An attempt to submit my FizzBizz solution


Solutions/session02/FizzBizz.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/python3
+for num in range(1,101,1):
```
PythonCHB: good, solid solution.

But you are building up the output string -- do you need to check for %3 and %5 twice?

PR #47, opened 2015-10-20
added broken cartalk puzzle7.  RFC

HasBob:
My function finds a couple instances of words containing three sequential pairs of double letters, but it does not find them all.


students/bob/session02/cartalk/tripleletters.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/python
+
+def triple():
+    '''Find all words with 3 double letters in a row.  From,
+    http://greenteapress.com/thinkpython/html/thinkpython010.html
+    http://www.cartalk.com/content/puzzlers  '''
+    fin = open('./words.txt')  ## word list to parse
+    for word in fin:
```
PythonCHB: this loops through all the lines in the file, assigning the name word to each line as it goes.

students/bob/session02/cartalk/tripleletters.py
```diff
@@ -0,0 +1,24 @@
...
+        word = fin.readline()
```
PythonCHB: and this reads a line from the file -- so you are actually skipping every other line...

PythonCHB:
NOTE: be careful about adding files that big -- over a MB!

students/bob/session02/cartalk/tripleletters.py
```diff
@@ -0,0 +1,24 @@
...
+        word = word.strip()
+        char = 0
+        length = len(word)
+        while (char <= length - 6) and (length >=6):
```
PythonCHB: this looks like it should work, but perhaps be a bit cleaner:

you should be able to use a for loop.

and the nested ifs could be done with "and"

PythonCHB:
oh -- and I didn't expect you to do the exercises -- but this is a good one :-)

-Chris

PR #49, opened 2015-10-21
FizzBuzz example

maxmac12:
Finished FizzBuzz example. Took me forever to realize that I was looking at the wrong branch for this pull request.


students/maxmaccamy/session02/fizzbuzz.py
```diff
@@ -0,0 +1,16 @@
+def fizzbuzz():
+    for i in range(100):
+        i += 1
```
PythonCHB: See if you can find a trick with range() so you don't need to add the 1 here.

PR #50, opened 2015-10-21
merge pull request for session02

tiffanyku:


students/tiffanyku/session_02/fizzbuzz.py
```diff
@@ -0,0 +1,11 @@
+n = 0
```
PythonCHB: no need to initialize n -- the for loop will do that

students/tiffanyku/session_02/grid_printer.py
```diff
@@ -0,0 +1,40 @@
...
+
+	grid = (row * 2 + top)
+
+	print(grid)
+
+
+one_var(11)
+
+
+def two_vars(w, h):
```
PythonCHB: nice!

PR #51, opened 2015-10-21
Lab assignments for session 02

HasMikeW:


students/mike_waddle/session_02/FizzBuzz.py
```diff
@@ -0,0 +1,12 @@
+for i in range(1,101):
+    if i % 15 == 0:
```
PythonCHB: nice catch on the 15 -- why check for 3 and 5 separately?

PR #54, opened 2015-10-21
session2 checkin

mdpitts2000:
check homework please


students/martypitts/session02/fizbuzz.py
```diff
@@ -0,0 +1,37 @@
+result = 1
+
+# end print_grid1
+
+# multiple of 3
+def mult3check(numin):
```
PythonCHB: nice job

PythonCHB:
Looks good -- I'll commit

PR #55, opened 2015-10-21
slicing lab + updates

HasBob:


students/bob/session02/series.py
```diff
@@ -0,0 +1,18 @@
...
+
+def fibonacci(n):
+    """Print n fibonnaci numbers"""
+    n = input('How many Fibonacci numbers to print? ')
+    fib=[0] # where we'll store the sequence list
+    x,y,z=0,1,1
+    for i in range(n-1):
+        fib.append(z)
+        z = x + y
+        x , y = y , z
```
PythonCHB: nice use of python multiple assignment

PythonCHB:
you've added a .DS_Store file -- not a big deal, but you want to be careful with:

``git add *``

You can also add a .gitignore file in your repo, and put .DS_store in there -- google a bit and you'll find recommended .gitignore files for OS-X and Python.

but don't add your .gitignore to the repo -- we all may have different needs...

students/bob/session03/slicinglab.py
```diff
@@ -0,0 +1,19 @@
+# week 3 lab
+
+lest = [0:21]
+
+def flip(lest):
+    return(lest[-1]+lest[1:-2]+lest[0])
```
PythonCHB: this will work with strings, but may not with other sequences. try:

flip(tuple(range(10))

students/bob/session03/slicinglab.py
```diff
@@ -0,0 +1,19 @@
...
+
+def skip(lest):
+    return(lest[::2])
+
+def d4(lest):
+    return(lest[4:-5:2])
```
PythonCHB: I think you want -4 here, not -5....

students/bob/session03/slicinglab.py
```diff
@@ -0,0 +1,19 @@
...
+
+def reverse(lest):
+    return(lest[::-1])
+
+def switch3rd(lest):
+    third = len(lest)/3
+    return(lest[third*2:third*3]+lest[third*0:third*1]+lest[thrid*1:third*2])
```
PythonCHB: third*3 should be the end -- plain old ``:``

and third*0 is 0 - so also a plain ``:``

PR #56, opened 2015-10-23
Pull request for 10.23.15 - rot13 and slicing homework submitted

JosephVC:


students/JosephC/Session_03/rot13.py
```diff
@@ -0,0 +1,42 @@
...
+# run the encryption/decryption code on each symbol in the message string
+for symbol in message:
+    if symbol in LETTERS:
+        #get the encrypted or decrypted number for this symbol
+        num = LETTERS.find(symbol) #get the number of the symbol (we need this to slice the message)
+        if mode == 'encrypt':
+            num = num + key
+        elif mode == 'decrypt':
+             num = num - key
+        #handle the wrap-around if num is larger than the length of LETTERS or less than 0; because the entire cipher is based on the letters in the alphabet, you can't have a key larger than the number of letters in the alphabet, you need to subtract or add the quantity of the key to 26 (number of letters in the alphabet); a key of zero does nothing, and a key greater than 26 simply wraps around the cipher
```
PythonCHB: you want to keep your comments (and code) to 80 characters or so per line...

students/JosephC/Session_03/rot13.py
```diff
@@ -0,0 +1,42 @@
...
+mode = 'decrypt' #set to 'encrypt' or 'decrypt'
+
+#every possible symbol that can be encrypted
+LETTERS = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+
+#stores the encrypted/decrypted form of the message
+translated = ''
+
+#capitalize the string in message
+message = message.upper()
```
PythonCHB: you should be abel to do it while preserving the capitalization.

The trick is to use teh number values of teh letters:

see ord(c)
and chr(i)

but you got all the logic right.

students/JosephC/Session_03/slicing.py
```diff
@@ -0,0 +1,34 @@
...
+    seq[::-1]
+
+sequence4('something')
+
+def sequence5(seq):
+# return a sequence with the middle third, then last third, then the first third in the new order
+
+    i = len(seq) //3
+    if 1 % 3:
+        print( seq[i*2:i*3+1] + seq[:i] + seq[i:i*3] )
```
PythonCHB: looks good.

Though you generally want to return the string, rather than printing it.

PR #58, opened 2015-10-25
Committing the mailroom project and updated the list lab

JosephVC:


students/JosephC/Session_03/list_lab.py
```diff
@@ -0,0 +1,64 @@
...
+    Add another fruit to the beginning of the list using insert() and display the list.
+    Display all the fruits that begin with “P”, using a for loop.
+
+
+"""
+def froots():
+    #   Create a list that contains “Apples”, “Pears”, “Oranges” and “Peaches” and display the list.
+    fruits = ['Apples', 'Pears', 'Oranges']
+
+    print("Our initial fruits are: " + fruits)
```
PythonCHB: this doesn't work -- you are trying to add a list and string together. however, this will:

print("Our initial fruits are: ", fruits)

students/JosephC/Session_03/list_lab.py
```diff
@@ -0,0 +1,64 @@
...
+
+    print('\n')
```
PythonCHB: print() by itself, will print an empty newline.

students/JosephC/Session_03/list_lab.py
```diff
@@ -0,0 +1,64 @@
...
+    #    Ask the user for another fruit and add it to the end of the list and display the list.
+    moar_fruits = input("This diet is terrible.  Add more fruits: ")
+
+    fruits.append(moar_fruits)
+
+    print('\n')
+    #    Ask the user for a number and display the number back to the user and the fruit corresponding to that number (on a 1-is-first basis)
+    #fruit_num = input("We need to number some fruit.  Enter a number to bring up the fruit in the corresponding location: ")
+
+    #for foo in fruits[0:int(fruit_num)]:
```
PythonCHB: we're just looking for a single fruit here -- so no need for a loop.

students/JosephC/Session_03/list_lab.py
```diff
@@ -0,0 +1,64 @@
...
+    print("Nope.  More.")    
+    fruits.insert(0, 'devil fruit')
+    print(fruits)
+
+    print('\n')
+    #   Display all the fruits that begin with “P”, using a for loop.
+    # NEEDS WORK    
+    print("Now we have to find only the fruits with P.")
+    for items in fruits:
+        if items[0] == 'p' or 'P':
```
PythonCHB: this will do ("p" or "P") first , which will always return "p". So you need to do:

items[0] == 'p' or items[0] == 'P'

or, I usually prefer:

items[0].lower() == 'p'

students/JosephC/Session_03/list_lab.py
```diff
@@ -0,0 +1,64 @@
...
+    #    Add another fruit to the beginning of the list using insert() and display the list.
+    print("Nope.  More.")    
+    fruits.insert(0, 'devil fruit')
+    print(fruits)
+
+    print('\n')
+    #   Display all the fruits that begin with “P”, using a for loop.
+    # NEEDS WORK    
+    print("Now we have to find only the fruits with P.")
+    for items in fruits:
```
PythonCHB: the usual convension is to use a singular name:

for item in fruits:

as "item" gets assigned to a single value each time through the loop.

students/JosephC/Session_03/mailroom.py
```diff
@@ -0,0 +1,13 @@
+#Mailroom
```
PythonCHB: It's a start!

PR #59, opened 2015-10-25
Corrections to list_lab and rot13

JosephVC:


students/JosephC/Session_03/list_lab.py
```diff
@@ -28,10 +28,11 @@ def froots():
 
     print('\n')
     #    Ask the user for a number and display the number back to the user and the fruit corresponding to that number (on a 1-is-first basis)
-    #fruit_num = input("We need to number some fruit.  Enter a number to bring up the fruit in the corresponding location: ")
+    fruit_num = input("We need to number some fruit.  Enter a number to bring up the fruit in the corresponding location: ")
 
-    #for foo in fruits[0:int(fruit_num)]:
-     #   print("Your number " + fruit_num + "brought up " + fruits[int(fruit_num)])
+    for foo in fruits[0:int(fruit_num)]:
```
PythonCHB: I was going for just the one number, not counting up -- but you get the idea.

Also, lit index form 0, but people don't usually think that way, so I was intending you to subtract one so they would get what they expected.

students/JosephC/Session_03/list_lab.py
```diff
@@ -41,18 +42,17 @@ def froots():
 
     print('\n')
     #    Add another fruit to the beginning of the list using insert() and display the list.
-    print("Nope.  More.")    
-    fruits.insert(0, 'devil fruit')
+    print("Nope.  More fruits.")    
+    fruits.insert(0, 'pineapple')
```
PythonCHB: I kind of liked "devil fruit" more :-)

students/JosephC/Session_03/list_lab.py
```diff
@@ -41,18 +42,17 @@ def froots():
...
 
     print('\n')
     #   Display all the fruits that begin with “P”, using a for loop.
-    # NEEDS WORK    
     print("Now we have to find only the fruits with P.")
     for items in fruits:
-        if items[0] == 'p' or 'P':
-            print(fruits)
-            break
+        if items.startswith('p') or items.startswith('P'):
```
PythonCHB: bingo!

students/JosephC/Session_03/mailroom.py
```diff
@@ -1,13 +1,59 @@
...
+We have a list of donors and the amounts they have donated. The user will be prompted to enter a donor name or simply generate a list of current donors.  
+If the name entered is not on the original list, then the user will enter that name and the amount donated, these values then being appended to the main donor list. 
+There is also the option of sending a Thank You email to a particular donor.
 
-def mailroom()
+'''
+#   define the donor list
+def donor_List():
+        
+        JohnGalt = [0]
```
PythonCHB: the donor names aren't code -- they are data -- so you want to store them in a data structure (lie list, hint hint), rather than use them as variable names. Much easy to add or remove one!

students/JosephC/Session_03/mailroom.py
```diff
@@ -1,13 +1,59 @@
...
+        donors = [JohnGalt, sparklyLestat, BiffBezos, GilliamWates ]
+#   a function to print the list of donors
+def report():
+    print(donors)
+#   create a new donor (to be used if a given name is not already on the list of donors)
+def create_New_Donor():
+     new_Donor = []
+     new_Name = input("Please enter the name of this new donor.")    
+     new_Donor.append(new_Name)
+     donors.append(new_Donor)
```
PythonCHB: you'll have the donation amount in the list, but not the name -- see previous comment.

students/JosephC/Session_03/slicing.py
```diff
@@ -2,33 +2,43 @@
 
 def sequence1(seq):
 #return a sequence with the first and last items exchanged
-
-    print (seq[-1:] + seq[1:-1]+seq[:0])    
+    seq = str(seq)
```
PythonCHB: why make it a string? this should work, as is, with any sequence.

PythonCHB:
By the way, no need to do a pull request for not-quite completed work, unless you want feedback at that stage.

JosephVC:
I do, as I feel I'm making the mailroom assignment more complicated than it
has to be.   :/

students/JosephC/Session_03/mailroom.py
JosephVC: So I will have a list of amount, but not names, in a single list?  Then how would I track the amounts with the names?

JosephVC: OK, I just remembered I can have a list of tuples rather than a whole slew of variables with lists attached to them

PythonCHB:
good -- then perfect. Feel free to ask specific questions in the comment...

students/JosephC/Session_03/mailroom.py
PythonCHB: Bingo! that's what I'd home you'd arrive at :)

And remember that you can put anything in a tuple -- including, say a list...., in addition to the name.

JosephVC: Now I'm having trouble having the names the user types in (to see what donor to email) be recognized in the donor list.  

The latter is formatted as follows:
donors = [ 
          ('John Galt', 0), 
          ('Sparkle Lestat', 10000), 
          ('Hef Bezos', 500, 500), 
          ('Gilliam Wates', 1000000, 1000000, 1000000) 
          ]

PythonCHB: That should work, the zeroth element is the names, so something like:

for donor in donors:
    if donor[0] == the_string_the_user_typed_in

Note, however, that your using tuple, which means you can't add a new donation to an existing donor. maybe try:

donors = [
...
('Hef Bezos', [500, 500]), 
....

so item 1 of the tuple is a list -- which you can append() to.

-CHB

JosephVC: I formatted my tuple of donors like the above, and when I try to do something like:

 >>>"John Galt" in donors

I get "False" returned.  Confused.  I started this exercise thinking I had things fairly well down, but apparently not . . .

codedragon: It is because the item you are referring to in the list is actually ('John Galt', 0), so 
('John Galt', 0) in donor 
should give you True. To do it this way, you have to loop through the list and ask each tuple if 'John Galt' is there.

students/JosephC/Session_03/mailroom.py
```diff
@@ -1,13 +1,63 @@
...
+'''
+We have a list of donors and the amounts they have donated. The user will be prompted to enter a donor name or simply generate a list of current donors.  
+If the name entered is not on the original list, then the user will enter that name and the amount donated, these values then being appended to the main donor list. 
+There is also the option of sending a Thank You email to a particular donor.
 
-def mailroom()
+'''
 
-    def report()
+donors = [
```
PythonCHB: nice -- this is what  came up with for a data structure.

students/JosephC/Session_03/mailroom.py
```diff
@@ -1,13 +1,59 @@
...
+        GilliamWates = [1000000, 1000000, 1000000]
+        donors = [JohnGalt, sparklyLestat, BiffBezos, GilliamWates ]
+#   a function to print the list of donors
+def report():
+    print(donors)
+#   create a new donor (to be used if a given name is not already on the list of donors)
+def create_New_Donor():
+     new_Donor = []
+     new_Name = input("Please enter the name of this new donor.")    
+     new_Donor.append(new_Name)
```
PythonCHB: this could be a touch more compact -- don't create an empty list, then:

new_Donor = [new_Name]

or even:

donors.append([new_Name])

students/JosephC/Session_03/mailroom.py
```diff
@@ -1,13 +1,59 @@
...
+        report()
+    #   append to the donors list if the name is not there by creating a new list and appending it to the original donors list  
+    elif name not in donors:
+        New_Donor()       
 
-    def thankYou():
-        name = input("Enter the full name of the recipient: "
+    
+#   main function 
+def mail_Room():
```
PythonCHB: This is all good -- now try to use a while loop to so the user will keep getting a prompt back untill they select "q" or quit or something....you go the core ideas though.

PR #61, opened 2015-10-26
Session3 Homework ROT13 and Mailroom

ryanmorin:
Rot13 complete but Mailroom is still being fine tuned

### DIFF
--- a/students/ryan_morin/session_02/series.py
+++ b/students/ryan_morin/session_02/series.py
@@ -1,40 +1,59 @@
-__author__ = 'Ryan'
+__author__ = 'ryan.morin'
 
 def fibonacci(n):
     """
-    :param n: number
-    :return: a list of nth value in the fibonacci series
+    :param n: An interger that describes the element of the fibonacci sequence to be returned
+    :return: The number that corresponds to the sequence
     """
-    fib_lst = [0,1]
-    for i in range(n):
-        fib_new = int(fib_lst[i]) + int(fib_lst[i+1])
-        fib_lst.append(fib_new)
-    return fib_lst[(len(fib_lst)-2)]
+    if n <= 1:
+        return 0
+    elif n == 2:
+        return 1
+    else:
+        return fibonacci(n-2) + fibonacci(n-1)
 
 def lucas(n):
     """
-    :param n: number
-    :return: a list of the nth value in the lucas series
+    :param n: An interger that describes the element of the lucas sequence to be returned
+    :return: The number that corresponds to the sequence
     """
-    lucas_lst = [2,1]
-    for i in range(n):
-        lucas_new = int(lucas_lst[i]) + int(lucas_lst[i+1])
-        lucas_lst.append(lucas_new)
-    return lucas_lst[(len(lucas_lst)-2)]
+    if n < 1:
+        return 0
+    elif n == 1:
+        return 2
+    elif n == 2:
+        return 1
+    else:
+        return lucas(n-2) + lucas(n-1)
 
 def sum_series(elmnt, op1=0, op2=1):
     """
-    :param elmnt: number
-    :param op1:  optional variable sequence start value - default equal to zero
-    :param op2: optional variable sequence start value - default equal to one
-    :return: return the nth value in a series depending on the optional parameters
+    :param elmnt: An interger that describes the element of the specific sequence to be returned
+    :param op1: first number in the series
+    :param op2: second number in the series
+    :return: The number that corresponds to the sequence
     """
     lst = [op1, op2]
-    for i in range(elmnt):
-        num_new = int(lst[i]) + int(lst[i+1])
-        lst.append(num_new)
-    return lst[(len(lst)-2)]
+    if op1 == 2 and op2 == 1:
+        return lucas(elmnt)
+    elif op1 == 0 and op2 == 1:
+        return fibonacci(elmnt)
+    else:
+        if elmnt == 0:
+            return 0
+        elif elmnt == 1:
+            return op1
+        else:
+            for i in range(elmnt):
+                temp = lst[i+1] + lst[i]
+                lst.append(temp)
+        return lst[elmnt-1]
 
-assert sum_series(5)
-assert sum_series(5,2,1)
-assert sum_series(5,4,8)
+assert (sum_series(0,0,1)) == 0
+assert (sum_series(3,0,1)) == 1
+assert (sum_series(3,2,1)) == 3
+assert (sum_series(3,2,5)) == 7
+
+
+
+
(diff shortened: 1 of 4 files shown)


students/ryan_morin/session_02/series.py
```diff
@@ -1,40 +1,59 @@
-__author__ = 'Ryan'
+__author__ = 'ryan.morin'
 
 def fibonacci(n):
     """
-    :param n: number
-    :return: a list of nth value in the fibonacci series
+    :param n: An interger that describes the element of the fibonacci sequence to be returned
+    :return: The number that corresponds to the sequence
```
PythonCHB: love the docstrings!

students/ryan_morin/session_02/series.py
```diff
@@ -1,40 +1,59 @@
...
-    for i in range(elmnt):
-        num_new = int(lst[i]) + int(lst[i+1])
-        lst.append(num_new)
-    return lst[(len(lst)-2)]
-
-assert sum_series(5)
-assert sum_series(5,2,1)
-assert sum_series(5,4,8)
\ No newline at end of file
+    if op1 == 2 and op2 == 1:
```
PythonCHB: why not just use this function here?

students/ryan_morin/session_02/series.py
```diff
@@ -1,40 +1,59 @@
...
+        return lucas(elmnt)
+    elif op1 == 0 and op2 == 1:
+        return fibonacci(elmnt)
+    else:
+        if elmnt == 0:
+            return 0
+        elif elmnt == 1:
+            return op1
+        else:
```
PythonCHB: why not use the recursive method you used before?

But you get it -- feel free to move on :-)

students/ryan_morin/session_03/mailroom.py
```diff
@@ -0,0 +1,105 @@
...
+
+def donor_list(dl):
+    """
+    Prints a list of names
+    :param dl: List of existing donors and amounts donated
+    :return: Print formatted list of names from list dl
+    """
+    print ('\n')
+    print ("The list of current donors:")
+    for d in dl:
```
PythonCHB: why not use your in_dl function here?

students/ryan_morin/session_03/mailroom.py
```diff
@@ -0,0 +1,105 @@
...
+    The function prints a formatted list of information pertaining to donors
+    :param dl: List of existing donors and amounts donated
+    :return: Prints a formatted list of names, number of donations, donation total and average donation
+    """
+    total = 0
+    print('\n')
+    for donor in dl:
+        doner_name = donor[0]
+        num_don = (len(donor)-1)
+        total = sum(donor[1:])
```
PythonCHB: nice find with sum()

students/ryan_morin/session_03/mailroom.py
```diff
@@ -0,0 +1,105 @@
...
+        avg_don = total//num_don
```
PythonCHB: do you want integer division here? -- it's your call, but you might want cents...

students/ryan_morin/session_03/mailroom.py
```diff
@@ -0,0 +1,105 @@
...
+        print ("Name: {:15} | Number of Donations: {:5} | Donation Total ${:>5} | Average Donation ${:>5}".format(
```
PythonCHB: nice format string

students/ryan_morin/session_03/mailroom.py
```diff
@@ -0,0 +1,105 @@
...
+    print ("\n")
+    return menu()
+
+def menu():
+    """
+    Prompts user and calls one of two functions based on response
+    """
+    print('\n')
+    action = input('Choose from the following actions: "Send a Thank You" or "Create a Report": ')
+    if action == 'Send a Thank You':
```
PythonCHB: you probably don't want to make your use type all that -- maybe just a numbered list of options, so they can type '1' or '2', or let them use just one letter.

students/ryan_morin/session_03/mailroom.py
```diff
@@ -0,0 +1,105 @@
...
+        print('\n')
+        return('Good bye')
+
+
+def thank_you():
+    """
+    The function prompts the user for two inputs, confirms one input is number and updates list dl
+    :return: prints an output to the console
+    """
+    response = ""
```
PythonCHB: no reason to "initialize" these -- it just gets tossed out when you assign them later.

students/ryan_morin/session_03/mailroom.py
```diff
@@ -0,0 +1,105 @@
...
+    amount = ""
+    while response != 'Quit':
+        response = input('Please provide a full name or type "Quit" to exit: ')
+        if response == 'Quit':
+            print('\n')
+            print ("...exiting program")
```
PythonCHB: if you call menu(), you're not exiting the program -- nit picky detail.

students/ryan_morin/session_03/mailroom.py
```diff
@@ -0,0 +1,105 @@
...
+            menu()
+        elif response.lower() == 'list':
+            donor_list(dl)
+        else:
+            while is_number(amount) == False:
+                amount = input("Please provide a donation amount: ")
+            if response not in in_dl(dl):
```
PythonCHB: maybe "in_dl" is not the next name for that function.....

students/ryan_morin/session_03/mailroom.py
```diff
@@ -0,0 +1,105 @@
...
+                dl.append(response)
+            for i in dl:
+                if response == i[0]:
+                    i.append(amount)
+            print('\n')
+            print ("Thanks {} for your generous ${:,} donation".format(response, int(amount)))
+            print('\n')
+
+if __name__ == '__main__':
+    print (menu())
```
PythonCHB: I didn't try to run it -- but overall looking good.

students/ryan_morin/session_03/rot13.py
```diff
@@ -0,0 +1,20 @@
+def rot13(input_strng):
+    rotted = ''
+    for i in input_strng:
+        ch = ord(i)
+        if ch >= ord('a') and ch <= ord('z'):
```
PythonCHB: maybe check for between a and m, then between m and z as one step. one less if....

also, between a and m you do the same thing as between A and M -- so you could combine those.

students/ryan_morin/session_03/rot13.py
```diff
@@ -0,0 +1,20 @@
...
+                ch += 13
+        elif ch >= ord('A') and ch <= ord('Z'):
+            if ch > ord('M'):
+                ch -= 13
+            else:
+                ch += 13
+        rotted += chr(ch)
+    return rotted
+
+if __name__ == '__main__':
```
PythonCHB: looks good!

PR #62, opened 2015-10-26
added rot13

HasBob:


students/bob/session03/mailroom.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/python
+
+# mailroom.py - automate thank-you note generation
+
+#  data structure that holds a list of donors & history of the amounts they've paid
+
+# notes from reading:
+#   start using raw_input in place of input.  input allows for code execution
```
PythonCHB: sorry -- that's an old py2 thing -- in py3 input() is what raw_input was in py2

students/bob/session03/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+
+donors = {'alice':[1,2,3],'bob':[10,20,30],'carl':[11,12,13],'dan':[44,66],'earl':[2]}
```
PythonCHB: nice!

students/bob/session03/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+    else:
+        return(mailroom())
+
+def send():
+    amount = 'temp' # variable to enter the while loop
+    ##del amount # need to ensure the varaible doesn't exist for later processing
+    donor = input("Enter donor's full name: ")
+    if donor not in donors: # if the name isn't in the dict
+        donors[donor] = []  # create it
+    while type(amount) != float:
```
PythonCHB: I'd usually do a ``while True`` here, then break when you get a valid result.

students/bob/session03/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+        try:
+           amount = input('Please enter the donation amount: $')
+           amount = float(amount)
+        except ValueError:
+            print('Float or int, buddy')
+
+    ##amount = input('Please enter the dollar amount: ')  ## this ruins everything until we ensure it's int or float.
+    ##print(amount,type(amount))
+    ##while type(amount) != (int or float):
```
PythonCHB: you found another way, but this wasn't doing what you wanted -- ``(int or float)`` is always going to evaluate to ``int``.

students/bob/session03/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+    donors[donor] += [amount] # add the amount to the donor's list
+    print('''
+        Dear {}, 
+        Project Arcturus couldn't have succeeded without your ${}.
+
+        Drop me a line if you're on the East Coast.
+        Hank Scorpio
+
+        '''.format(donor,amount))
+    print(mailroom())
```
PythonCHB: probably better to keep your interactive loop clean -- run it by itself, and don't have mailroom() return what you want to print.

students/bob/session03/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+
+def report():
+    print('\nreports:  COMING EVENTUALLY')
```
PythonCHB: :-)

students/bob/session03/rot13.py
```diff
@@ -0,0 +1,39 @@
...
+            create an empty string called 'code', append characters to 
+            it as they're decrptyed (temporarily stored in 'a')
+    '''
+    code = '' # decoded text aggregator
+    for i in x:
+        if i in 'abcdefghijklm':
+            a = ord(i)+13
+            a = chr(a)
+            code += a
+        elif i in 'nopqrstuvwxyz':
```
PythonCHB: this will work fine, but maybe better to work with the ordinals -- something like:

```
if   ord(a) <= ord(i) <= ord(m)
```

better yet, pre-compute the ord(a).

or do all of a-z then wrap around if it went too large.. % is helpful there.

PR #63, opened 2015-10-26
Session 3 Labs

iampei:
Slicing and List Lab


iampei:
don't have any luck on the mailroom code....definitely need some help here. Thank you.

PythonCHB:
I know it's annoying to have everyone else's stuff cluttering up your machine, but that's just how git works. the whole repo is the project. If I merge this, it will remove everyone else's stuff from the class repo -- we don't want that.

This may be a bit tricky to clean up -- probably easier to make a new fork of the class repo, then put your new stuff in it and try another pull request.

PR #64, opened 2015-10-26
rot13 assert problem

gabrielx52:
I got the rot13 function to work with the translate method but the asserts won't work, they are throwing a flag even when its working properly. I'm not sure why. Any insight would be appreciated. Thanks

### DIFF
--- a/students/gabriel_meringolo/session_02/boolLab.py
+++ b/students/gabriel_meringolo/session_02/boolLab.py
@@ -1,2 +1,22 @@
+''' When squirrels get together for a party,
+    they like to have cigars. A squirrel party
+    is successful when the number of cigars is
+    between 40 and 60, inclusive. Unless it is
+    the weekend, in which case there is no upper
+    bound on the number of cigars. Return True if
+    the party with the given values is successful,
+    or False otherwise. '''
+
 def cigar_party(cigars, is_weekend):
     return (cigars >= 40 and (cigars <= 60 or is_weekend))
+
+
+''' The squirrels in Palo Alto spend most of the day
+    playing. In particular, they play if the temperature
+    is between 60 and 90 (inclusive). Unless it is summer,
+    then the upper limit is 100 instead of 90. Given an int
+    temperature and a boolean is_summer, return True if the
+    squirrels play and False otherwise. '''
+
+def squirrel_play(temp, is_summer):
+    return temp >= 60 and (temp <= 90 or (temp <= 100 and is_summer))
--- a/students/gabriel_meringolo/session_02/boolLab.py
+++ b/students/gabriel_meringolo/session_02/boolLab.py
@@ -1,0 +1,2 @@
+def cigar_party(cigars, is_weekend):
+    return (cigars >= 40 and (cigars <= 60 or is_weekend))
--- a/students/gabriel_meringolo/session_03/mail_menu.py
+++ b/students/gabriel_meringolo/session_03/mail_menu.py
@@ -58,7 +58,6 @@ def mail_menu(x):
     print("--------------------------------------------\a")
     menu = input("1: Send a 'Thank You' Letter to Donor\n2: Create Donor report\n3: Quit\n>")
     if menu == "1":
-        #name = input("")
         thank_you(x)
     if menu == "2":
         report(x)
@@ -130,7 +129,6 @@ def donor_check(x):
     :param y: donor name
     :return:
     '''
-    #new_don = ""
     y = input("\nEnter Donor name\n> ")
     if y.lower() == "quit":
         quit_mr()
@@ -138,10 +136,7 @@ def donor_check(x):
         print("\nDonor Found")
         return y
     elif y not in flat_list(x):
-        #print(new_don, "rad")
-        #new_don = y
-        #print(new_don, "radder")
-        add_donor(x,y)
+        return add_donor(x,y)
 
 
 def add_donate(x, y):
--- a/students/gabriel_meringolo/session_03/mail_menu.py
+++ b/students/gabriel_meringolo/session_03/mail_menu.py
@@ -1,9 +1,12 @@
 donors = [["Graham", 1, 5, 10], ["Eric", 4, 6], ["Terry", 5, 5, 5], ["John", 20], ["Michael", 10, 10]]
+doncopy = donors[:]
+
 
 
 def get_two(item):
     '''returns second item in list'''
     return item[1]
+
 
 
 def report(x): #compiles the donor list descending by total donations
@@ -23,6 +26,7 @@ def report(x): #compiles the donor list descending by total donations
     print("\n")
 
 
+
 def mail_head(x):
     '''
     prints mailroom greeting
@@ -34,7 +38,7 @@ def mail_head(x):
     print("|           V 1.0            |")
     print("#----------------------------#\n\n")
     input("Press enter to continue.\a\n")
-    print("Welcome to the Python Endowment Fund mailroom.\n\n")
+    print("Welcome to the Python Endowment Fund mailroom.")
     mail_menu(x)
 
 
@@ -50,7 +54,7 @@ def mail_menu(x):
     :param x: donor database list
     :return:
     '''
-    print("Please make a selection from the menu below.")
+    print("\nPlease make a selection from the menu below.")
     print("--------------------------------------------\a")
     menu = input("1: Send a 'Thank You' Letter to Donor\n2: Create Donor report\n3: Quit\n>")
     if menu == "1":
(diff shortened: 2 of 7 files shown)


gabrielx52:
I got the student file lab working properly, the only issue I encountered was that Java was in the list 4 times with one occurrence being uppercase. So its in my display twice, one for the upper and once for the lower. I wasn't sure the best way of joining the two without affecting the format of other languages in uppercase C/C#...
I'm sure i could compare them with .lower() But I erred on the side of concise code and skipped it.

students/gabriel_meringolo/session_03/rot13.py
```diff
@@ -0,0 +1,17 @@
+
+
+
+def rot13(i):
+    one = "aAbBcCdDeEfFgGhHiIjJkKlLmMnNoOpPqQrRsStTuUvVwWxXyYzZ"
+    another = "nNoOpPqQrRsStTuUvVwWxXyYzZaAbBcCdDeEfFgGhHiIjJkKlLmM"
+    transtab = str.maketrans(one, another)
+    print(i.translate(transtab))
```
PythonCHB: you aren't returning the result....

students/gabriel_meringolo/session_03/rot13.py
```diff
@@ -0,0 +1,17 @@
...
+
+if __name__ == "__main__":
+    print("main")
+    assert rot13("a") == "n"
```
PythonCHB: since you didn't return it will return None, so this is:

```
assert None == "n"
```

but nice find of .translate() and maketrans() !

students/gabriel_meringolo/session_04/dict_lab.py
```diff
@@ -0,0 +1,77 @@
...
+print(py_set)
+
+#Create a frozenset with the letters in ‘marathon’
+m_set = frozenset(["m","a","r","a","t","h","o","n"])
+print(m_set)
+
+
+#display the union and intersection of the two sets.
+print("union: ", py_set.union(m_set))
+print("intersection: ", py_set.intersection(m_set))
```
PythonCHB: all looks good.

students/gabriel_meringolo/session_04/file_lab.py
```diff
@@ -0,0 +1,18 @@
...
+lan_str = "" # enpty string
+lan_list = [] # empty list
+
+for i in studs:
+    lan_list += (i[i.find(":") + 1:].replace(",","").split())
+    """seperating names from languages/ removing comas
+    and creating a list from the text doc with only the languages"""
+lan_list.sort() # sorting list by alpha
+
+for ii in lan_list:
```
PythonCHB: try to use meaningful names for loo variables -- `i` and the like is fine if it's a simple integer, like an index, but it's much easier ot read the code if you use a beter name -- maybe somethingn like:

```
for lang in lan_list:
   ....

File: students/gabriel_meringolo/session_04/file_lab.py [hunk 5]
```diff
@@ -0,0 +1,18 @@
...
+    if ii not in lan_str:
```
PythonCHB: this will prevent duplicates -- but maybe use a set so you won't have duplicates in the first place. (though case is tricky there...)

students/gabriel_meringolo/session_02/boolLab.py
```diff
@@ -0,0 +1,22 @@
...
+    they like to have cigars. A squirrel party
+    is successful when the number of cigars is
+    between 40 and 60, inclusive. Unless it is
+    the weekend, in which case there is no upper
+    bound on the number of cigars. Return True if
+    the party with the given values is successful,
+    or False otherwise. '''
+
+def cigar_party(cigars, is_weekend):
+    return (cigars >= 40 and (cigars <= 60 or is_weekend))
```
PythonCHB: nice!

students/gabriel_meringolo/session_03/mail_menu.py
```diff
@@ -0,0 +1,196 @@
+donors = [["Graham", 1, 5, 10], ["Eric", 4, 6], ["Terry", 5, 5, 5], ["John", 20], ["Michael", 10, 10]]
+
+
+def get_two(item):
```
PythonCHB: why have a function for this? and why do you commonly want the second item?

students/gabriel_meringolo/session_03/mail_menu.py
```diff
@@ -0,0 +1,196 @@
+donors = [["Graham", 1, 5, 10], ["Eric", 4, 6], ["Terry", 5, 5, 5], ["John", 20], ["Michael", 10, 10]]
```
PythonCHB: remember that you can nest objects -- so the list of donations could be self contained:
``.... ("Graham", [1,5,10]), ....``

a little cleaner and easier to access the list of donations as a single object

students/gabriel_meringolo/session_03/mail_menu.py
```diff
@@ -0,0 +1,196 @@
...
+def report(x):
+    '''
+    compiles the donor list descending by total donations
+    :param x: donor database list
+    :return:
+    '''
+    d_list = []
+    for i in x:
+        d_list.append(list(((i[0], sum(i[1:]), len(i) - 1, (sum(i[1:]) // (len(i) - 1))))))
+    d_list = sorted(d_list,key=get_two, reverse=True)
```
PythonCHB: ah -- now I see what the get_two is about -- well done.

students/gabriel_meringolo/session_03/mail_menu.py
```diff
@@ -0,0 +1,196 @@
...
+    print("\nDonors              Total Amt           Total Donations        Average Amt        ")
+    print("------------------------------------------------------------------------------\n")
+    for i in d_list:
+        donor = str(i[0])
+        totes = str(i[1])
+        donx = str(i[2])
+        ave = str(i[3])
+        print("{}".format(donor) , " " * (20 - len(str(i[0]))) , "${}".format(totes) ,
```
PythonCHB: you can use format spcecifiers to make colums, something like:

```
In [18]: "{:20}{:10}{:20}".format("this", "that", "the other")
Out[18]: 'this                that      the other  
```

students/gabriel_meringolo/session_03/mail_menu.py
```diff
@@ -0,0 +1,196 @@
...
+              " " * 18 , "{}".format(donx) , " " * 22 + "${}".format(ave),)
+    print("\n")
+
+
+
+def mail_head(x):
+    '''
+    prints mailroom greeting
+    :param x: donor database list
```
PythonCHB: use meaningful names -- 'x' could be anything.

students/gabriel_meringolo/session_03/mail_menu.py
```diff
@@ -0,0 +1,196 @@
...
+    :return:
+    '''
+    print("\nPlease make a selection from the menu below.")
+    print("--------------------------------------------\a")
+    menu = input("1: Send a 'Thank You' Letter to Donor\n2: Create Donor report\n3: Quit\n> ")
+    if menu == "1":
+        thank_you(x)
+    if menu == "2":
+        report(x)
+        mail_menu(x)
```
PythonCHB: recursive function calls for this will work -- but not ideal.

try a while loop that you break out of when the user says they want to quit.

students/gabriel_meringolo/session_03/mail_menu.py
```diff
@@ -0,0 +1,196 @@
...
+    '''
+    checks for donor on list
+    :param x: donor database list
+    :param y: donor name
+    :return:
+    '''
+    y = input("\nEnter Donor name\n> ")
+    if y.lower() == "quit":
+        quit_mr()
+    if y in flat_list(x):
```
PythonCHB: this is making a while copy, just  to search for something -- a little more structure in the donor-list would make this unnecessary.

students/gabriel_meringolo/session_03/mail_menu.py
```diff
@@ -0,0 +1,196 @@
...
+
+def add_donate(x, y):
+    '''
+    adds a donation
+    :param x: donor database list
+    :param y: donor name
+    :return:
+    '''
+    z = input("\nEnter donation amount\n> ")
+    if num_check(x, y, z) == True:
```
PythonCHB: you eant to us ``is True`` rather than ``== True``. but in this case, just plain:

``if num_check:``

will do it -- remember "truthiness"

students/gabriel_meringolo/session_03/mail_menu.py
```diff
@@ -0,0 +1,196 @@
...
+def num_check(x, y, z):
+    '''
+    checking for float or int
+    :param x: donor database list
+    :param y: donor
+    :param z: donation amount
+    :return:
+    '''
+    try:
+        float(z) or int(z)
```
PythonCHB: this will do the wrong thing if z is "0.0"

why not actually use the conversion? -- i.e. return float(z)

also, every integer is a float, too, so no need to check both.

PR #65, opened 2015-10-26
revised list_lab and wrote rot13

robalford:


students/robalford/session03/mailroom.py
```diff
@@ -0,0 +1,94 @@
...
+    return donor_list
+
+
+def select_command():
+    command = input("""Commands:
+Enter 'Thank you' to send a thank you message.
+Enter 'Report' to create a report.
+Enter 'Home' to return to this screen.
+Enter 'Quit' to exit My Donation Manager.""")
+    while command.lower() == 'thank you':
```
PythonCHB: you can just use a if here. and one while loop around the whole thing.

PythonCHB: then you wouldn't need the recursive calls.

students/robalford/session03/mailroom.py
```diff
@@ -0,0 +1,94 @@
...
+            print(donor[0])
+        full_name = prompt_user("Enter the donor's full name or type 'list' to see all donors.")
+
+    donation_amount = prompt_user('Enter the donation amount.')
+
+    while donation_amount:
+        try:
+            donation_amount = float(donation_amount)
+            break
+        except:
```
PythonCHB: yup!

students/robalford/session03/mailroom.py
```diff
@@ -0,0 +1,94 @@
...
+    Sincerely,
+
+    Your favorite charity """.format(full_name, donation_amount)
+
+    print(thank_you)
+    return select_command()
+
+
+def create_report():
+    global donors
```
PythonCHB: donors is mutable -- no need for global here.

students/robalford/session03/rot13.py
```diff
@@ -0,0 +1,38 @@
...
+    all_the_characters = list(string.ascii_lowercase)
+    new_string = ''
+    upper = False
+    for i in str:
+        # preserve capitalization.
+        if i.isupper():
+            upper = True
+            i = i.lower()
+        # preserve punctuation
+        if i in string.punctuation:
```
PythonCHB: why not just pass through if it's not a letter?

students/robalford/session03/rot13.py
```diff
@@ -0,0 +1,38 @@
...
+            new_string += i
+        # preserve whitespace
+        elif i in string.whitespace:
+            new_string += i
+        else:
+            location = all_the_characters.index(i)
```
PythonCHB: this is a pretty slow operation -- better to work with the ordinal values -- ord('c').

students/robalford/session03/rot13.py
```diff
@@ -0,0 +1,38 @@
+import string
+
+
+def rot13(str):
+    all_the_characters = list(string.ascii_lowercase)
+    new_string = ''
+    upper = False
+    for i in str:
+        # preserve capitalization.
+        if i.isupper():
```
PythonCHB: interesting -- I was thinking just work with the upper case letters separately -- but this is kind of slick...

students/robalford/session04/dict_lab.py
```diff
@@ -0,0 +1,53 @@
...
+
+# sets
+
+divisible_by2 = set()
+divisible_by3 = set()
+divisible_by4 = set()
+
+for i in range(20):
+    if i % 2 == 0:
+        divisible_by2.update([i])
```
PythonCHB: might as well use add() here -- for adding a single item

students/robalford/session04/files_lab.py
```diff
@@ -0,0 +1,30 @@
+#
+students = open('/Users/robalford/Desktop/python_practice/python_class/IntroPython2015/Examples/students.txt')
```
PythonCHB: you can do relative paths here:

"../../Examples/students.txt'"

students/robalford/session04/files_lab.py
```diff
@@ -0,0 +1,30 @@
...
+
+languages = []
+for line in students:
+    line = line.split(': ')
+    line.pop(0)
+    if line:
+        line[0] = line[0][:-1]
+        line[0] = line[0].split()
+    languages += line
+students.close
```
PythonCHB: you need to call the close function: 

students.close()

students/robalford/session04/files_lab.py
```diff
@@ -0,0 +1,30 @@
+#
+students = open('/Users/robalford/Desktop/python_practice/python_class/IntroPython2015/Examples/students.txt')
+
+languages = []
+for line in students:
+    line = line.split(': ')
+    line.pop(0)
+    if line:
+        line[0] = line[0][:-1]
```
PythonCHB: is this getting rid of the newline? you can use str.strip() for that.

and why line[0]? shouldn't it be the only item left? -- so you could do:

line = line.split(':')[1]

and get what's after the colon.

students/robalford/session04/files_lab.py
```diff
@@ -0,0 +1,30 @@
...
+    languages += line
+students.close
+languages.pop(0)
+
+language_list = []
+language_frequency = {}
+for student in languages:
+    for language in student:
+        language = language.lower()
+        if language[-1] == ',':
```
PythonCHB: maybe better to have stripped those out earlier.

PR #66, opened 2015-10-27
Week 3 homework

elaewin:
I'm not sure what I'm doing wrong with the if/else statements in the get_name function in mailroom.py, but it always writes "exit" to the list of donors.


students/erica_winberry/session03/mailroom.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # Include Donor Name, total donated, number of donations and average donation amount as values in each row.
+    # Using string formatting, format the output rows as nicely as possible. The end 
+    # result should be tabular (values in each column should align with those above and below)
+    # After printing this report, return to the original prompt.
+    # At any point, the user should be able to quit their current task and return to the original prompt.
+    # From the original prompt, the user should be able to quit the script cleanly
+
+
+def get_name(l):
```
PythonCHB: you really want to use a more meaningful name than "l" ....

students/erica_winberry/session03/mailroom.py
```diff
@@ -0,0 +1,152 @@
...
+            break
+        # If the user types ‘list’, show them a list of the donor names and re-prompt
+        elif donor_name.lower() == "list":
+            print("The current list of donors is: ")
+            for i in l:
+                print(i[0])
+        else:
+            # If the user types a name in the list, use it.
+            count = 0
+            for i in l:
```
PythonCHB: enumerate would help here:

```
for i, donor in l:
    if i[0] == name:
        break
```

PythonCHB:
the while loop in get_name looks OK to me -- did you fix it?

students/erica_winberry/session03/mailroom.py
```diff
@@ -0,0 +1,152 @@
...
+def intro():
+    print("\nMAILROOM OPTIONS")
+    print("1 - Send a Thank You")
+    print("2 - Create a Report")
+    print("\n(Type 'exit' at any time.)\n")
+
+
+def mailroom():
+    intro()
+    while True:
```
PythonCHB: nicely done interactive prompt loop.

students/erica_winberry/session03/mailroom.py
```diff
@@ -0,0 +1,152 @@
...
+            intro()
+
+
+# Sending a Thank You
+def thanks():
+    donor_name = get_name(donors)
+    if donor_name == "exit":
+        return print("Returning to main menu.\n")
+    donation = get_amount(donor_name, donors)
+    if donation is not 0:
```
PythonCHB: you want to use !=, rather than is not for comparing to zero.

students/erica_winberry/session03/mailroom.py
```diff
@@ -0,0 +1,152 @@
...
+def write_letter(donor,amount):
+    # Write a thank you letter for a charitable contribution. Make sure that "charity" and 
+    # "signature" are correct for the organization sending the letter.
+    if amount == 0:
+        return ("\nNo new donations. No letter will be written.\n")
+    else:
+        charity = "Carter Home for Retired Superheroes"
+        closing = "Sincerely"
+        signed = "S.A. Carter"
+        print("\nDear {},".format(donor))
```
PythonCHB: you probably want to make one big string out of this -- so you can do something other than print it to the screen.

students/erica_winberry/session03/rot13.py
```diff
@@ -0,0 +1,40 @@
...
+
+# As usual, add your new file to your local clone right away.
+# Make commits early and often and include commit messages that are descriptive and concise.
+
+
+
+def rot13_translate(l):
+    base_alpha = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+    plus13 = "NOPQRSTUVWXYZABCDEFGHIJKLMnopqrstuvwxyzabcdefghijklm"
+    switcher = str.maketrans(base_alpha, plus13)
```
PythonCHB: nice find!

students/erica_winberry/session03/rot13.py
```diff
@@ -0,0 +1,40 @@
...
+    new = l.translate(switcher)
+    print(new)
+
+def rot13_encode(l):
```
PythonCHB: did you notice that these are the same ?

students/erica_winberry/session03/string_formatting_lab.py
```diff
@@ -0,0 +1,25 @@
...
+# numbers = (1,2,3)
+# print("The first three numbers are {:d}, {:d}, {:d}".format(*numbers))
+
+def format_lab():
+    numbers = (1,2,3,4,5,6,7,8,9)
+    l = len(numbers)
+    print("There are " + str(l) + " numbers in the list.")
+    i = int(input("How many of the numbers would you like to print? "))
+    if i <= l: 
+        result = "\nThe first " + str(i) + " numbers are: " + ("{:d}, " * (i)).format(*numbers)
```
PythonCHB: nice!

students/erica_winberry/session04/file_lab.py
```diff
@@ -0,0 +1,46 @@
+import os
+
+os.chdir("../../../Examples")
+student_data = []
+f = open("students.txt", "r")
+for line in f:
+    line = line.split(":")
+    student, langs = line
+    if "\n" in langs:
```
PythonCHB: the \n will always be at the end -- I usually call .strip() on the line first thing.

students/erica_winberry/session04/file_lab.py
```diff
@@ -0,0 +1,46 @@
...
+        langs = langs.replace("\n", "")
+    if "," in langs:
```
PythonCHB: you don't have to check first -- if you call replace, and it's not there, it will do nothing.

students/erica_winberry/session04/file_lab.py
```diff
@@ -0,0 +1,46 @@
...
+    student_data.append(langs)
+f.close()
+
+for i in student_data:
+    for t in i:
+        if t == "and":
+            i.remove(t)
+        if t == "":
+            i.remove(t)
+        if "," in t:
```
PythonCHB: if you cleaned out the commas and "and" before you split it up --  it would have been easier.

PR #67, opened 2015-10-28
session3

mdpitts2000:


students/martypitts/session03/rot13.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+'''
+
+# Encode the values using rot13 algroythm encrypt
+def rot13(seq):
+    encodestr = []
+    asciistr = []
+    for i in seq:
+        asciival = int(ord(i))
```
PythonCHB: ord(i) will always be an int....

students/martypitts/session03/rot13.py
```diff
@@ -0,0 +1,118 @@
...
+        if asciival > 64 and asciival < 78:
```
PythonCHB: you can write this as:

```
if 64 < asciival < 78:
```

a little more compact.

students/martypitts/session03/rot13.py
```diff
@@ -0,0 +1,118 @@
...
+            rot13val = asciival + int(13)
```
PythonCHB: 13 is defiantly always an int!

students/martypitts/session03/rot13.py
```diff
@@ -0,0 +1,118 @@
...
+        elif asciival > 109 and asciival < 123:
+            rot13val = asciival - int(13)
+        elif asciival > 31 and asciival < 48:
+            rot13val = asciival + int(13)
+        asciistr.append(asciival)
+        encodestr.append(rot13val)
+    # print("The ascii string=", asciistr)
+    # print("The rot13 encoded string=", encodestr)
+
+    return encodestr
```
PythonCHB: this is now a list -- you'll want to join it into one string:

"".join(encodestr)

students/martypitts/session03/rot13.py
```diff
@@ -0,0 +1,118 @@
...
+
+# Print out the encoded message
+def rot13chardump(seq):
+    encrypted = []
+    for i in seq:
+        charval = chr(i)
+        encrypted.append(charval)
+    # print("The encrpyted message = ", encrypted)
+    return encrypted
```
PythonCHB: now you probably want to "".join(encrypted)

students/martypitts/session03/rot13.py
```diff
@@ -0,0 +1,118 @@
...
+
+# Decode the values using rot13 algroythm decrypt
+def rot13inv(seq):
```
PythonCHB: did you notice that it's the same :-)

PR #69, opened 2015-10-30
trigrams.py

robalford:
Can you confirm that I'm on the right track with trigrams.py so far before I add more text processing? Also I don't know why another student's work is showing up in my pull request below.


PythonCHB:
absolutely on the right track -- I'll comment more later.

Not sure what's up with the other students' files -- I"ll look into that, too.

but did you pull from upstrem, and then push to origin? that should so it right.

codedragon:
I confirmed locally that merging this would not change anything in AustinScara's directory. Not sure why that was showing up, but it will be fine, so I'm going to merge.

PR #70, opened 2015-10-31
Files for Session 4; hoping trigram is on a right track

JosephVC:


students/JosephC/Session_04/trigram.py
```diff
@@ -0,0 +1,28 @@
...
+
+a = {'I may': 'I', 'may I': 'wish', 'I wish': ('I', 'I'), 'wish I': ('may', 'might')}
+
+for i in a:
+    print("the {}th item is {}".format(i, a))
+
+
+#create variables to hold each line read from sherlock and then to hold the random choice of line
+line = f.readline('sherlock.txt')
+randomline = random.choice(line)
```
PythonCHB: you really don't want to be picking random lines -- you want create a full list of all the words in the text, build up the trigrams from that -- then randomly pick from the trigrams to make new text.

PythonCHB:
It looks like you deleted other student's directories form your repo -- I can see why you'd want to do that, but if I merge this, then their stuff will get deleted form the main repo.

JosephVC:
So what would I need to do 're: other repos?

PR #72, opened 2015-11-01
Not sure what route to go..

gabrielx52:
I've gotten it to print out the trigrams in its simplest form but I'm stuck on how to go from here. 
I've attempted this problem from a few different angles- stripping off two words at a time then looping through the whole txt and if found stripping off the next word... lots of converting strings to lists and back and more than a few hours with enumerate.
Any suggestions for the best mode of attack whole be most appreciated.

### DIFF
--- a/students/gabriel_meringolo/session_03/rot13.py
+++ b/students/gabriel_meringolo/session_03/rot13.py
@@ -1,11 +1,11 @@
 
 
 
-def rot13(i):
+def rot13(text_to_trans):
     one = "aAbBcCdDeEfFgGhHiIjJkKlLmMnNoOpPqQrRsStTuUvVwWxXyYzZ"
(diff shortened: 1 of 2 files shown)


PythonCHB:
A few ideas.

first step is to rad the file and get a list of words - you are doing that, but you'll need to a bit of text processing to strip off the header, maybe deal with punctuations, etc.

Second step is to build up the trigrams -- so you want to loop through the words, as you are, but you do want to access the three words in a row, so you do need indexes, so enumerate not that helpful:

for i in range(len(list_of_words) - 2):
    pair = words[i:i+2]
    follower = words[i+2]

in this loop, you need to build up a data structure to map work pairs to a list of words tha follwo that pair.

hints:

 - You can use a tuple of strings as a key in a dict.
 dict.setdefault() is helpful here.

Third step is to randomly select form that dict to build up your text.

HTH,

-Chris

PR #73, opened 2015-11-01
added session3 and 4

iampei:


students/pei/session03/list_lab.py
```diff
@@ -0,0 +1,72 @@
+#Series 1
+#Create a list that contains “Apples”, “Pears”, “Oranges” and “Peaches”.
+list = ['Apples', 'Pears', 'Oranges', 'Peaches']
+#Display the list
```
codedragon: list is a keyword, and you should not use it as a variable name. Use my_list or fruit_list or just fruit. To add something to the end of a list, just use 

my_list.append(newFruit).

students/pei/session03/list_lab.py
```diff
@@ -0,0 +1,72 @@
...
+newList  = ['Strawberries'] + list
+#display the list.
+print (newList)
+#Add another fruit to the beginning of the list using insert()
+newList.insert (0, 'Pomegranates')
+#display the list
+print (newList)
+#Display all the fruits that begin with “P”, using a for loop
+print ([fruit for fruit in newList if (fruit[0] is 'P')])
+
```
codedragon: You discovered list comprehensions! Very cool. We will talk about comprehensions on Tuesday.

students/pei/session03/list_lab.py
```diff
@@ -0,0 +1,72 @@
...
+        newList.remove(fruit)
+# for each fruit in the list (making the fruit all lowercase)
+# Display the list
+print ([x.lower() for x in newList])
+
+#Series 4
+newList = ['Pomegranates', 'Strawberries', 'Apples', 'Pears', 'Oranges', 'Peaches']
+#Make a copy of the list
+import copy
+copiedList = copy.copy(newList)
```
codedragon: You can also make a copy by either doing:
copiedList = newList[:]
or 
copiedList = list(newList)

students/pei/session03/mailroom.py
```diff
@@ -0,0 +1,80 @@
...
+    newList = [ ]
+    for key, value in donorNames.items():
+        newList.append([key, sum(value), len(value), sum(value)/len(value)])
+    sorted(newList, key = lambda donor: donor [1])
+        #print (sorted(donorNames.items()))
+
+elif tORr ==3:
+    print ('The program ended')
+
+
```
codedragon: The key to getting the structure to work (so that the user is always returned to the main prompt, and can navigate from adding donors to making a report) is to use small functions and have a main loop. If you aren't sure what I mean by that, ask me about it at office hours next week or take a look at Chris's solution.

students/pei/session03/strFormat.py
```diff
@@ -0,0 +1,18 @@
...
+'''
+#'{:.2f}'.format(123.4567)
+#'{:0>3d}'.format (2)
+#'{:.0e}'.format(10000)
+
+'file_{:0>3d}, {:.2f}, {:.0e}'.format (2, 123.4567, 10000)
+# Rewrite: "the first 3 numbers are: {:d}, {:d}, {:d}".format(1,2,3) 
+# to take an arbitrary number of values
+t = (1,2,3)
+"the first 3 numbers are: {:d}, {:d}, {:d}".format(*t)
```
codedragon: The idea was to print all of the numbers, along with the total where the 3 was. So, for example, "the first 5 numbers are 1, 2, 3, 4, 5". You may want to take a look at Chris's solution. As you saw when I taught class, most programmers don't have this stuff memorized, but you do need to be able to figure out how to do it, if you need to.

students/pei/session04/session4labs.py
```diff
@@ -0,0 +1,84 @@
...
+f = open('students.txt')
+student_data = f.read()
+list = student_data.split('\n')
+#print (list)
+
+for i in list:
+    print (i.rpartition(':')[-1])
+f.close()
+
+
```
codedragon: Nice use of rpartition. Would be great to get it into an actual list, without duplicates. :)

codedragon:
You are figuring stuff out! It takes time and practice, but you are getting the hang of it.

PR #76, opened 2015-11-03
added kata14 code

iampei:
Did the home work...but my code didn't get to break out of the while loop, it raised an error: 

TypeError: object of type 'NoneType' has no len()

for this line of code: new += " " + ''.join(random.choice(grouping2.get(strLastTwo)))

So I added an exception, but then it never got to break out of the while loop. I don't know if that's alright...

Pei

### DIFF
--- a/students/pei/session04/kata14.py
+++ b/students/pei/session04/kata14.py
@@ -1,0 +1,57 @@
+txt = open('sherlocksmall.txt')
+txt_data = txt.read().replace("\n", " ").strip().split(' ')
+#txt_data = txt.read()
+#grab the first 3, put them in the list, increment by 1 and grab the next 3
+#to create a dictionary
+grouping1 =[ ]
+grouping2 ={ } 
+for n in range(len(txt_data)-2):
+    del grouping1 [:]
+    grouping1.append(txt_data[n] + " " + txt_data[n+1])
+    grouping1.append(txt_data[n+2]) 
+    #print (grouping1[0])
+    if grouping1[0] not in grouping2:
+        #grouping2[(grouplist1[0])] = (grouping1[1]) 
+        grouping2[grouping1[0]] = [grouping1[1]]
+    else:
+        grouping2[grouping1[0]].append(grouping1[1])
+
+print (grouping2)
+
+#generate new sentance using the dict
+#def generate (strText):
+import random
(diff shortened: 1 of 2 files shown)


PythonCHB:
well, it should end at SOME point. :-)

I think it's trying to do a random.choice() call on None -- it has to figure out the length first to make a random choice -- and None has no length.

so you want to figure out how the None got there -- maybe at the end of the input?

I"ll take a quick look and see if I can see it.

students/pei/session04/kata14.py
```diff
@@ -0,0 +1,57 @@
+txt = open('sherlocksmall.txt')
+txt_data = txt.read().replace("\n", " ").strip().split(' ')
+#txt_data = txt.read()
+#grab the first 3, put them in the list, increment by 1 and grab the next 3
+#to create a dictionary
+grouping1 =[ ]
+grouping2 ={ } 
+for n in range(len(txt_data)-2):
+    del grouping1 [:]
```
PythonCHB: might as well just make a new list...

students/pei/session04/kata14.py
```diff
@@ -0,0 +1,57 @@
...
+#generate new sentance using the dict
+#def generate (strText):
+import random
+new = "he was"
+strLastTwo = new
+#strLastTwo in grouping2
+while strLastTwo in grouping2:
+    strLastTwo = ' '.join(new.split()[-2:])
+    try: 
+        new += " " + ''.join(random.choice(grouping2.get(strLastTwo)))
```
PythonCHB: I think maybe strLastTwo is not guaranteed to be in the dict -- so the get() call returns None.

students/pei/session04/kata14.py
```diff
@@ -0,0 +1,57 @@
...
+
+print (grouping2)
+
+#generate new sentance using the dict
+#def generate (strText):
+import random
+new = "he was"
+strLastTwo = new
+#strLastTwo in grouping2
+while strLastTwo in grouping2:
```
PythonCHB: this could indeed go on forever -- you need to terminate at some point!

maybe just arbitrarily do 100 words or so..

PythonCHB:
You are very, very close -- nice work!

PR #80, opened 2015-11-04
Sorry for the giant pull request

iancote:
It's been a bit since I updated.


PythonCHB:
It's a lot! --  so anything in particular you'd like reviewed?

Also -- those are some big text files -- I like the idea of trying out the trigrams on different books, but better not add them all to the repo. Then everyoen will get them in their repos, too. Please git rm them, and then you can still keep copies in your working directory as untracked files.

-CHB

students/iancote/session04/file_copy.py
```diff
@@ -0,0 +1,20 @@
...
+
+import sys
+
+if len(sys.argv) == 3:
+    source, destination = sys.argv[1:3]
+    with open(source, 'rb') as s:
+        with open(destination, 'wb') as d:
+            while True:
+                buf = s.read(1024)
+                if len(buf) == 0:
```
PythonCHB: this could be simply: 

if buf

( a zero-length bytes object is "Falsey")

but otherwise perfect.

students/iancote/session04/files_lab.py
```diff
@@ -0,0 +1,20 @@
+languages = {}
+
+with open("../../../Examples/students.txt", 'r') as f:
+    for line in f:
+        line = line.rstrip()
```
PythonCHB: extra, as you will strip() the end of the line later anyway -- but no harm

students/iancote/session04/files_lab.py
```diff
@@ -0,0 +1,20 @@
...
+        student, lingos = line.split(":")
+        lingos = lingos.strip()
+        for l in lingos.split(" "):
```
PythonCHB: .split() without any arguments splits on all whitespace -- often the easiest way to go

students/iancote/session04/trigram.py
```diff
@@ -0,0 +1,70 @@
...
+
+    with open(filename, 'r') as f:
+        while True:
+            line = f.readline()
+            if not line or line.startswith(
+                                 '*** END OF THIS PROJECT GUTENBERG'):
+                break
+            if line.startswith('*** START OF THIS PROJECT GUTENBERG'):
+                continue
+            l += re.findall(r"[\w']+|[.,!?;]", line)
```
PythonCHB: I see you found regular expressions -- well done.

students/iancote/session04/trigram.py
```diff
@@ -0,0 +1,70 @@
...
+            for i in range(len(l) - 2):
+                try:
+                    if l[i + 2] not in db["{} {}".format(l[i], l[i + 1])]:
+                        db["{} {}".format(l[i], l[i + 1])].append(l[i + 2])
+                except KeyError:
+                    db["{} {}".format(l[i], l[i + 1])] = [l[i + 2]]
```
PythonCHB: nice!

PythonCHB:
Thanks for removing the extra files -- this looks great.

PR #81, opened 2015-11-04
Add kata14.py

ryanmorin:
My solution

### DIFF
--- a/students/ryan_morin/session_04/Lab.py
+++ b/students/ryan_morin/session_04/Lab.py
@@ -1,0 +1,40 @@
+__author__ = 'Ryan'
+
+food_prefs = {"name": u"Chris",
+              u"city": u"Seattle",
+              u"cake": u"chocolate",
+              u"fruit": u"mango",
+              u"salad": u"greek",
+              u"pasta": u"lasagna"}
+
+# Quesstion 1
+print ("{name} is from {city}, and he likes {cake} cake, {fruit}, {salad} salad and {pasta} pasta"
+       .format(**food_prefs))
+
+#Question #2 /#3
+c = {i:hex(i) for i in range (16)}
+print (c)
+
+fp_copy = food_prefs.copy()
+
+#Question #4
+d = {i:v.count("a") for i,v in food_prefs.items()}
+print (d)
+
+#Qestion#5
+e = {i for i in range(21)}
+s2 = e.copy()
+s3 = e.copy()
+s4 = e.copy()
+
+s2 = {(i/2) for i in e}
+s3 = {round((i/3),1) for i in e}
+s4 = {round((i/4),1) for i in e}
+
+print(s2,s3,s4)
+
+list_a = list(e)
+list_b = [2,3,4]
+
+oput = [[(x*y) for x in list_a] for y in list_b ]
+print (oput)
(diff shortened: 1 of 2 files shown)


students/ryan_morin/session_04/Lab.py
```diff
@@ -0,0 +1,40 @@
...
+
+#Qestion#5
+e = {i for i in range(21)}
+s2 = e.copy()
+s3 = e.copy()
+s4 = e.copy()
+
+s2 = {(i/2) for i in e}
+s3 = {round((i/3),1) for i in e}
+s4 = {round((i/4),1) for i in e}
```
PythonCHB: This doesn't quit get what i was looking for - the idea was to get teh number divisible by 4 -- so, 0,4,8,12,....

you can get it this way with some tweaking:

 s4 = {int(round((i/3),0))*3 for i in e}

and that's kind of cool, as it takes advantage of the set to remove duplicates.

though I was thinking more like: 

s3 = { i for i in e if not i%3}

students/ryan_morin/session_04/kata14.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    This function opens a text file.
+    :return: The full text file
+    """
+    orig_doc = open('c:/Users/Ryan/Sherlock.txt', 'r')
+    lines = orig_doc.readlines()
+    orig_doc.close()
+    return lines
+
+def splt(lst):
```
PythonCHB: keeping it as a list of lines adds a level of dat structures that don't really buy you anything. the individual lines are  layout artifact, not part of the meaning of the text -- so easier to put it all together into one big ordered list of words.

you could do that by using read() rather than readlines() when you read the file, or do it here with:

com_man_list.extend(wrd_lst)

students/ryan_morin/session_04/kata14.py
```diff
@@ -0,0 +1,86 @@
...
+    return (com_man_list)
+
+def str_loop():
+    """
+    The loop removes special characters from the words
+    :return: a list of words that has the special characters removed
+    """
+    chr_to_repl = ':;()/<>?"'
+    for line in splt(open_file()):
+        for wrd in line:
```
PythonCHB: see how this would be easier if it wasn't still split into lines?

students/ryan_morin/session_04/kata14.py
```diff
@@ -0,0 +1,86 @@
...
+    return (wrd_list)
+
+def dict_seed():
+    """
+    This function joins the words i and i+1 adds it to a dictionary and adds i+2 as the dictionary's vakue
+    :return: a dictionary with the words from the input file
+    """
+    l = str_loop()
+    for i in range(len(l)-2):
+        key_pair = (l[i])+" "+(l[i+1])
```
PythonCHB: or " ".join(l[i:i+2]) -- though when it' sonly two, it's a wash.

students/ryan_morin/session_04/kata14.py
```diff
@@ -0,0 +1,86 @@
...
+        val = (l[i+2])
+        if key_pair in d:
+            d.setdefault(key_pair, []).append(val)
```
PythonCHB: setdefault() is pretty nifty, eh?

PR #82, opened 2015-11-04
Trigram v 2.0

gabrielx52:
It looks like I was improving this as you were checking the old version. I feel like there might be a bit of redundancy in the formatting but besides that it should be good.


students/gabriel_meringolo/session_04/trigram.py
```diff
@@ -1,26 +1,28 @@
 import random
+import textwrap
```
PythonCHB: good find!

PythonCHB:
All looks good -- nice work.

PR #83, opened 2015-11-05
Session 05 HW

robalford:


students/robalford/session05/.cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
PythonCHB: this shouldn't have been added to git.

PythonCHB:
looks like a bunch of cache and temp files got added to git -- pytest does add those.

beware "git add *" ! -- you want to be careful about what you add. I usually do thinks like:

git add *.py *.txt

if I have a bunch to add.

you'll need to remove them with:

git rm __pycache__/*

etc....

students/robalford/session05/comprehensions.py
```diff
@@ -0,0 +1,46 @@
+# 1
+
+food_prefs = {"name": u"Chris",
+              u"city": u"Seattle",
+              u"cake": u"chocolate",
+              u"fruit": u"mango",
+              u"salad": u"greek",
+              u"pasta": u"lasagna"}
+
+# Is this the preferred (Pythonic) style for wrapping long lines of text?
```
PythonCHB: yes, this is common style. You can also use a triple quoted string:

person_likes_food = """{name} is from {city}, and he likes {cake} cake,
{fruit} fruit, {salad} salad and {pasta} pasta.""".format(**food_prefs)

though that will put in a newline.

students/robalford/session05/comprehensions.py
```diff
@@ -0,0 +1,46 @@
...
+# s2 = {number for number in range(21) if number % 2 == 0}
+# s3 = {number for number in range(21) if number % 3 == 0}
+# s4 = {number for number in range(21) if number % 4 == 0}
+
+# sets = []
+
+# for num in range(2, 5):
+#     sets.append({number for number in range(21) if number % num == 0})
+
+sets = [{number for number in range(21) if number % num == 0} for num in range(2, 5)]
```
PythonCHB: nice!

students/robalford/session05/test_coding_bat.py
```diff
@@ -0,0 +1,66 @@
+from coding_bat import string_splosion, cigar_party, count_evens
```
PythonCHB: you got the idea -- nice tests.

robalford:
Should I add .pyc to my .gitignore as well to make it explicit that these
kinds of files should never be tracked by git? Or are there any
circumstances where you would want to add them?

PythonCHB:
yes, adding *.pyc to .gitignore is a good idea.

cewing:
Just an FYI, github maintains a pretty sensible [Python .gitignore file](https://github.com/github/gitignore/blob/master/Python.gitignore).  When you start a new repository, it's an option to begin with one in place, and even for existing repos, adding it by pasting what you find at that link is quite a good starting place.

PythonCHB:
Thanks Cris,

I hadn't noticed that.

And I'll also point you to this repo:

https://github.com/github/gitignore

Which has comprehensive .gitignore files for all kinds of projects.

Chris

On Tuesday, November 10, 2015, Cris Ewing <notifications@github.com> wrote:

> Just an FYI, github maintains a pretty sensible Python .gitignore file
> <https://github.com/github/gitignore/blob/master/Python.gitignore>. When
> you start a new repository, it's an option to begin with one in place, and
> even for existing repos, adding it by pasting what you find at that link is
> quite a good starting place.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/UWPCE-PythonCert/IntroPython2015/pull/83#issuecomment-155576242>
> .
>


-- 
Christopher Barker, PhD

Python Language Consulting
  - Teaching
  - Scientific Software Development
  - Desktop GUI and Web Development
  - wxPython, numpy, scipy, Cython

PR #85, opened 2015-11-08
added exception lab

iampei:
Not sure why there were a lot of example file changes being tracked...


PythonCHB:
Those are all cache files left behind by pytest -- they should never have been added to git.

That's why will happen if you do a "git add *"
It's possible that I did that, ven though I should know better.

But I'll merge and clean it up.

students/pei/session04/exceptionLab.py
```diff
@@ -0,0 +1,10 @@
+safeStr = ""
```
PythonCHB: no need to initialize this

students/pei/session04/exceptionLab.py
```diff
@@ -0,0 +1,10 @@
...
+def safe_input (name):
```
PythonCHB: you aren't doing anything with name here.

you probably want to pass in "prompt", and then use it:

def safe_input(prompt):
...
        safeStr = input(prompt)

PR #88, opened 2015-11-09
Dict list lab V2.x

gabrielx52:
I completed the first part of this lab, but I think I'm overlooking something big or I may just be misunderstanding the instructions..
The code is simple and clean and gets the job done but I feel like I should be able to do it with dict comprehension instead of explicitly calling on the keys. That also leads me to believe that that would over complicate the code.. 

edit: I did the list comp and the dict comp parts as well. I'm not really sure how to create a dict using list comprehension. I ended up using zip on two lists that I created with list comprehension. hopefully I didn't miss the point of the assignment.

### DIFF
--- a/students/gabriel_meringolo/session_05/dict_list_lab2.py
+++ b/students/gabriel_meringolo/session_05/dict_list_lab2.py
@@ -1,8 +1,8 @@
-#food_prefs = {"name": "Chris", "city": "Seattle", "cake": "chocolate",
-#              "fruit": "mango", "salad": "caesar", "pasta": "lasagna"}
+food_prefs = {"name": "Chris", "city": "Seattle", "cake": "chocolate",
+              "fruit": "mango", "salad": "caesar", "pasta": "lasagna"}
 
-#print("{name} is from {city} he likes {cake} cake,\n{fruit} smoothies, extra dressing \
-#on his {salad}\nsalads and always asks for seconds of {pasta}.".format(**food_prefs))
+print("{name} is from {city} he likes {cake} cake,\n{fruit} smoothies, extra dressing \
+on his {salad}\nsalads and always asks for seconds of {pasta}.".format(**food_prefs))
 
 
 #Using a list comprehension, build a dictionary of numbers from zero to fifteen and the hexadecimal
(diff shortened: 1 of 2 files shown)


students/gabriel_meringolo/session_05/dict_list_lab2.py
```diff
@@ -0,0 +1,18 @@
...
+              "fruit": "mango", "salad": "caesar", "pasta": "lasagna"}
+
+print("{name} is from {city} he likes {cake} cake,\n{fruit} smoothies, extra dressing \
+on his {salad}\nsalads and always asks for seconds of {pasta}.".format(**food_prefs))
+
+
+#Using a list comprehension, build a dictionary of numbers from zero to fifteen and the hexadecimal
+#equivalent (string is fine). (the hex() function gives you the hexidecimal representation of a number.)
+
+numbs = [x for x in range(16)]
```
PythonCHB: this works, but: list(range(16)) is more efficient and clearer.
Also -- a range object is already an iterable sequence, so no need to make it a list at all.

students/gabriel_meringolo/session_05/dict_list_lab2.py
```diff
@@ -0,0 +1,18 @@
...
+hex_numbs = [hex(x) for x in numbs]
+print({x: y for x, y in zip(numbs, hex_numbs)})
```
PythonCHB: yup -- zip() is the way to go if you have two lists -- but do you need to make the list of hex_numbs ahead of time?

See if you can simply call hex() inside the dict comp.

students/gabriel_meringolo/session_05/list_comp_lab.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+#print('first' in dict_comprehension) #fear? False  the comp puts the keys in caps -this doesn't exist as a key
+#print('FIRST' in dict_comprehension) #True                                        -this does!
+#print(len(dict_of_weapons)) #5
+#print(len(dict_comprehension)) #4
+
+
+def count_evens(nums):
+    return len([x for x in nums if x % 2 == 0])
```
PythonCHB: yup!

PythonCHB:
mostly looks good -- you're getting it.

PR #89, opened 2015-11-09
improved exception lab

iampei:


students/pei/session04/exceptionLab.py
```diff
@@ -1,10 +1,9 @@
-safeStr = ""
-def safe_input (name):
+def safe_input (prompt):
     try:
-        safeStr = input("please provide your first name:")
+        safeStr = input(prompt)
         return"Hi {0}, I hope you are having a great day!".format(safeStr)
```
PythonCHB: you probably want to return safeStr here -- so the callign function can do whatever it wants with it. But you got the Exception Handling right, which was the point.

Also -- safeStr is not a PEP8-compliant name -- use safe_str:

https://www.python.org/dev/peps/pep-0008/#naming-conventions

it's a good habit to get into for making your code more readable to other pythonistas

PR #91, opened 2015-11-10
Catching up on homework

elaewin:
I'm still trying to get error handling to work properly in my mailroom program from last week. I keep getting a type error.


PythonCHB:
really nice commit messages!

a lot here -- I'll look at mailroom2, but let me know if there is anything else in particular you want me to look at.

students/erica_winberry/session04/mailroom2.py
```diff
@@ -0,0 +1,188 @@
...
+template – rather than building up a big string in parts.
+
+THEN:
+Add Exception handling to mailroom
+and add some tests
+and list (and dict, and set) comprehensions...
+"""
+
+donors = [
+    {"name": "Carol Danvers", "donations": [25.00, 100.00]},
```
PythonCHB: you may want to put the whole thing in a dict, rather than a list -- using the name , or some variation of it, like all lower case with no spaces, or.... as a key, so you can quickly grab the one you want without having to search through them all.

students/erica_winberry/session04/mailroom2.py
```diff
@@ -0,0 +1,188 @@
...
+    else:
+        print("Returning to main menu.\n")
+
+
+def report(report_source):
+    # Print a list of your donor, sorted by total historical donation amount.
+    print("{:^80}".format("DONOR REPORT\n"))
+    print("{:>20}{:>20}{:>20}{:>20}".format("Donor Name", "Total Amt. \
+Donated", "Times Donated", "Avg. Donation"))
+    print("{:>20}".format("------------------") * 4)
```
PythonCHB: nice formatting!

students/erica_winberry/session04/mailroom2.py
```diff
@@ -0,0 +1,188 @@
...
+    for donor_dictionary in report_source:
```
PythonCHB: I'd probably just call this "donor" -- you know it's a dict, and why type that much?

students/erica_winberry/session04/mailroom2.py
```diff
@@ -0,0 +1,188 @@
...
+        donor = donor_dictionary["name"]
+        total_amount = sum(donor_dictionary["donations"])
+        total_donations = len(donor_dictionary["donations"])
+        try:
+            avg_donation = total_amount / total_donations
+        except ZeroDivisionError as e:
+            return e
```
PythonCHB: you generally do'nt want to return the Exception object -- you want to handle the exception -- in this case, I'd say set the average_donation to 0.0 -- then that's it.

If you do something, but still need the error propogate, then you re-raise it:

raise e

students/erica_winberry/session04/mailroom2.py
```diff
@@ -0,0 +1,188 @@
...
+        new_donation = safe_input("Please enter the amount of the new \
+donation from " + donor + " to the nearest whole dollar: $")
+        # Verify that the amount is in fact a number, and re-prompt if it isn’t.
+        # Once an amount has been given, add that amount to the donation history of the selected user.
+        if new_donation is None:
+            break
+        elif new_donation.lower() == "exit":
+            break
+        elif new_donation.isnumeric():
+            new_donation = float(new_donation)
```
PythonCHB: here is a good place for "it's better to ask forgiveness than permission":

try:
    new_donation = float(new_donation)
except:
    handle the error -- probably continue

students/erica_winberry/session04/mailroom2.py
```diff
@@ -0,0 +1,188 @@
...
+                new_dict = {"name": donor_name, "donations": []}
+                donor_list.append(new_dict)
+                return donor_name
+
+
+def get_amount(donor, donations):
+    # Get the amount of a new donation for a thank you letter.
+    # Append a new donation to the list of donations made by an individual in list donationans.
+    new_donation = ""
+    while type(new_donation) is not float:
```
PythonCHB: a little strange to terminate the loop in a type check. how about:

new_donation = None
while new_donation is not None

or simply while not new_donation # any falsey value

or while True:

then terminate the loop some other way.

students/erica_winberry/session04/mailroom2.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+# Finally, use string formatting to compose an email thanking the donor for 
+# their generous donation. Print the email to the terminal and return to the original prompt.
+def write_letter(donor, amount):
+    # Write a thank you letter for a charitable contribution. Make sure that "charity" and 
+    # "signature" are correct for the organization sending the letter.
+    if amount == 0:
+        return ("\nNo new donations. No letter will be written.\n")
+    elif amount == "exit":
```
PythonCHB: you probably want to put this logic elsewhere -- why call this function if there is no need to use it?

students/erica_winberry/session04/mailroom2.py
```diff
@@ -0,0 +1,188 @@
...
+        new_letter = print("\nDear {},".format(donor) + "\nThank you \
+so much for your generous donation of ${:.2f} to \
+the \n{charity}.".format(amount, charity="Carter Home for Retired \
+Superheroes") + "\n{closing},".format(closing="Sincerely") +
+            "\n{signed}".format(signed="S.A. Carter"))
+        return new_letter
+
+
+def generate_letters(dictionary_of_donors):
+    for dictionaries in dictionary_of_donors:
```
PythonCHB: is dictionary_of_donors a dictionary? - I think it's a list of dicts...

personally, I avoid names that specify a type -- but if you are going to -- make sure it's correct! really confusing otherwise!

students/erica_winberry/session04/mailroom2.py
```diff
@@ -0,0 +1,188 @@
...
+you for your generous donation(s) totaling ${:.2f} to the \
+\n{charity}.".format(total_amount, charity="Carter Home for Retired \
+Superheroes") + "\n{closing},".format(closing="Sincerely") +
+            "\n{signed}".format(signed="S.A. Carter"))
+        with open(file_name, "w") as f:
+            f.write(new_letter)
+            print("Letter written for {}.".format(donor))
+
+if __name__ == "__main__":
+    while True:
```
PythonCHB: nice clean menu loop -- the other sub-menus can be this clean, too!

PythonCHB:
Nice unit tests for the coding bat exercises.

PythonCHB:
I can't merge this, because there are conflicts -- which is weird, that usually only happens when two different people edit the same file.

try doing teh merge on you maachine first:

git pull upstream master

then you may have some merge conflicts -- fix those conflicts, then:

git commit -a

git push origin master (or just git push)

then ping me to look again, by adding a comment to this PR.

PythonCHB:
Merge looks all clean now -- merging.

But let me know if you want anything in particular reviewed.

PR #92, opened 2015-11-11
assignments up to but not including mailroom

ericrosko:
I'm turning in the homework up to but not including the mailroom.  I've gotten a bit behind but I'll catch up quickly.  It is not a matter of me not being able to do it, and I'm not looking at the solutions so everything is my own work.  Everything is pretty straightforward and I did all the CodingBat problems yesterday.

All the readings and coding just ended up taking more time than I thought!

Thanks,
Eric


students/ericrosko/session02/fizzbuzz.py
```diff
@@ -0,0 +1,74 @@
...
+	You must have py.test installed from http://pytest.org
+	python3 -m pip install pytest
+Usage: 
+	py.test -v fizzbuzz.py
+	or
+	python3 fizzbuzz.py
+
+'''
+
+def create_array(n=100):
```
PythonCHB: if you want a list, rather than a range object (sometimes you do), just do:

list(range(n))

though in this case, I think a range object would be fine -- it acts almost like a tuple -- i.e. an immutable sequence.

students/ericrosko/session02/fizzbuzz.py
```diff
@@ -0,0 +1,74 @@
...
+	array = []
+	for i in range(1,n+1):
+		array.append(i)
+	return array
+
+def fizzbuzz(array=create_array()):
```
PythonCHB: this is the issue I talked about i class this week:

create_array() will be called when the function is defined -- then every time fizzbuzz is called, the same one will be used. in this case, you aren't mutating it, so it'll be fine, but be careful!

students/ericrosko/session02/fizzbuzz.py
```diff
@@ -0,0 +1,74 @@
...
+	return output
+	
+
+if __name__ == "__main__":
+	print(fizzbuzz())
+
+#
+# tests
+#
+def test_create_array():
```
PythonCHB: love the tests!

PythonCHB:
It looks like you are using tabs for indentation -- this will work, and is not TOO bad if you are consistent about it, but the standard is four spaces -- and you really, really want to do that -- trust me on this.

Make sure to set up your editor to use four spaces for indentation, and it should be able to auto-convert for you, too.

students/ericrosko/session02/series.py
```diff
@@ -0,0 +1,84 @@
...
+	second = 1
+	if Lucas:
+		first=2
+		second=1
+
+	if n == 0:
+		return first
+	elif n == 1:
+		return second
+	else:
```
PythonCHB: nice iterative solution -- you may want to try the recursive way, too (calling fibonacci from within fibonacci, just so you can see how that works -- it's a natural way to solve a problem like this.

PythonCHB:
I'm loving the unit tests! nice work.

PR #94, opened 2015-11-12
labs from class and mailroom tests/refactoring

robalford:


students/robalford/session03/mailroom.py
```diff
@@ -4,7 +4,12 @@
     'Ringo Starr': [1.00],
     'George Harrison': [1000.00, 500.00, 1586.78],
     'Yoko Ono': [275.50, 5.00]
-}
+    }
+
+
+def list_donors():
+    donor_list = [donor for donor in donors.keys()]
+    return '\n'.join(donor_list)
```
PythonCHB: no need to make alist first:

'\n'.join(donors.keys())

students/robalford/session03/mailroom.py
```diff
@@ -29,48 +34,63 @@ def write_email(donor, donation_amount):
...
+def add_new_donation(full_name, donation_amount):
+    # donors = get_donors()
     while True:
-        if full_name.lower() == 'list':
-            for donor in donors.keys():
-                print(donor)
-                break
-        elif full_name.lower() == 'home':
+        # check to see is user typed 'home'
+        if isinstance(donation_amount, str) and donation_amount.lower() == 'home':
```
PythonCHB: input always produces a string -- why do an isinstance() check?

students/robalford/session03/mailroom.py
```diff
@@ -29,48 +34,63 @@ def write_email(donor, donation_amount):
...
-        donors[full_name] = [donation_amount]
+    add_new_donation(full_name, donation_amount)
 
     print(write_email(full_name, donation_amount))
 
 
 def create_report():
-    print('Donor\t\t\tTotal\t\tNumber\t\tAverage')
+    donor_reports = []
     for donor in donors:
```
PythonCHB: good opportunity to do:

for donor, donations in donors.items():
    ....

then you don't need to keep pulling the donations list out.

students/robalford/session03/mailroom.py
```diff
@@ -29,48 +34,63 @@ def write_email(donor, donation_amount):
...
+        # report = '{}\t\t{:.2f}\t\t{}\t\t{:.2f}\n'.format(*donor_report)
+        donor_reports.append(donor_report)
+    return donor_reports
+
+
+def print_report():
+    donor_reports = create_report()
+    print('Donor\t\t\tTotal\t\tNumber\t\tAverage')
+    for report in donor_reports:
+        print('{}\t\t{:.2f}\t\t{}\t\t{:.2f}\n'.format(*report))
```
PythonCHB: nice formatting!

students/robalford/session03/test_mailroom.py
```diff
@@ -0,0 +1,39 @@
...
+    mailroom.add_new_donation('New Donor', 23.00)
+    report = mailroom.create_report()
+    assert ('New Donor', 23.0, 1, 23.0) in report
+    # verify average donation
+    assert report[0][1]/report[0][2] == report[0][3]
+
+
+def test_create_letter_files():
+    mailroom.create_letter_files()
+    assert os.path.isfile('John Lennon.txt')
```
PythonCHB: nice tests!

students/robalford/session06/kwargs_lab.py
```diff
@@ -0,0 +1,18 @@
...
+
+colors = ('black', 'white')
+
+print_colors(*colors)
+
+
+def print_arbitrary_colors(*args, **kwargs):
+    print(args, kwargs)
+
+print_arbitrary_colors('blue', 'black', fore_color='white')
```
PythonCHB: looks like you're getting the idea.

students/robalford/session06/lambda_keyword.py
```diff
@@ -0,0 +1,5 @@
+def function_builder(n):
+    functions = []
+    for i in range(n):
+        functions.append(lambda value, increment=i: value + increment)
```
PythonCHB: bingo!

PythonCHB:
there are confilcts in the PR -- which makes no sense, as you're the only one editing.

But try:

git pull upstream master
(there may be conflicts then)
fix the conflicts
git commit 
git push

then that should clear it up.

students/robalford/session06/trapz.py
```diff
@@ -0,0 +1,82 @@
+from math import sqrt
+from functools import partial
+
+
+def line(x):
+    return 5
+
```
codedragon: I think you may have wanted return x, otherwise, why bother asking for an argument? ;-)

students/robalford/session06/trapz.py
```diff
@@ -0,0 +1,82 @@
...
+    # return grid_spacing
+    list_of_values = [a + grid_spacing]
+    for i in range(num_of_grids-2):
+        list_of_values.append(list_of_values[-1] + grid_spacing)
+    # return list_of_values
+    list_of_values = [2*fun(i, **coef) for i in list_of_values]
+    list_of_values.append(fun(a, **coef)+fun(b, **coef))
+    return (b-a)/(2*num_of_grids) * sum(list_of_values)
+
+
```
codedragon: Well done. Another way to do the list of values is to increase grid_spacing in the loop, and then you can do a list comprehension: 
alt_list_values = [a + i * grid_spacing for i in range(num_of_grids + 1)]

students/robalford/session06/trapz.py
PythonCHB: actually, this is right -- the argument is to conform to the expected API: what is y for a given value of x? in this case, always a constant -- but the trapz function will call it by passing a value, so it needs to take an argument, whether it needs it or not.

codedragon: I know it is technically correct, but it is still a silly function! So, I
was trying to make a joke, but it was late. Anyway, it really should have a
comment or be called line_5 or something, since it is a specific line.

~m

On Mon, Nov 16, 2015 at 11:33 PM, Christopher H.Barker, PhD <
notifications@github.com> wrote:

> In students/robalford/session06/trapz.py
> <https://github.com/UWPCE-PythonCert/IntroPython2015/pull/94#discussion_r45029143>
> :
>
> > @@ -0,0 +1,82 @@
> > +from math import sqrt
> > +from functools import partial
> > +
> > +
> > +def line(x):
> > +    return 5
> > +
>
> actually, this is right -- the argument is to conform to the expected API:
> what is y for a given value of x? in this case, always a constant -- but
> the trapz function will call it by passing a value, so it needs to take an
> argument, whether it needs it or not.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/UWPCE-PythonCert/IntroPython2015/pull/94/files#r45029143>
> .
>



-- 
Maria Mckinley
Software Developer
www.mariakathryn.net
www.linkedin.com/in/mariamckinley


*NOTICE:* Due to Presidential Executive Orders still not rescinded by
President Obama, the National Security Agency may have read this email
without warning, warrant, or notice, and certainly without probable cause.
They may do this without any judicial or legislative oversight. Click here
to learn more about current US laws and your rights
<http://www.aclu.org/national-security/surveillance-privacy>.

robalford: no worries. i caught the joke :)

PR #95, opened 2015-11-13
added a few labs.

gabrielx52:
I"ve been struggling with the concept of lambda, I've been reading from a lot of different sources about it and finally got it but its still a bit fuzzy...
If you know of any blogs/books/podcasts that are geared toward a person with a non-tech background I'd be interested. Thanks.


PythonCHB:
This one's not bad:

http://www.secnetix.de/olli/Python/lambda_functions.hawk

All your reading may have made it harder -- there us a bunch of theory about lambda, particularly in other languages. But in Python, it is simply alternative syntax for defining a simple function. That's it, really.

students/gabriel_meringolo/session_04/update_mailroom.py
```diff
@@ -1,16 +1,16 @@
...
     print("\nDonors              Total Amt           Total Donations        Average Amt        ")
     print("------------------------------------------------------------------------------")
-    for i in donor_dict:
-        donor = i
-        donations = str((donors.get(i))).strip("[]")
-        total = sum(donors.get(i))
-        average = sum(donors.get(i))//len(donors.get(i))
+    for donor in donor_dict:
+        donations = str((donor_dict.get(donor))).strip("[]")
+        total = sum(donor_dict.get(donor))
```
PythonCHB: I would usually use plain old: donor_dict[donor] here. The difference is get() will return None (or the value you specify if the key is not in the dict, whereas regular old indexing will raise an exception.

students/gabriel_meringolo/session_04/update_mailroom.py
```diff
@@ -1,16 +1,16 @@
...
 
 def report(donor_dict):
     print("\nDonors              Total Amt           Total Donations        Average Amt        ")
     print("------------------------------------------------------------------------------")
-    for i in donor_dict:
-        donor = i
-        donations = str((donors.get(i))).strip("[]")
-        total = sum(donors.get(i))
-        average = sum(donors.get(i))//len(donors.get(i))
+    for donor in donor_dict:
```
PythonCHB: this wold be a good palce to do:

for donor, donations in donors.items():
    ....

then you get both at once, and don't have to access the list later.

students/gabriel_meringolo/session_04/update_mailroom.py
```diff
@@ -62,12 +62,11 @@ def donor_names(donor_dict):
...
     prints list of donor names
     :param donor_dict: donor list
-    :return: donors ini x
+    :return:
     '''
-    print("\n")
-    for i in donor_dict:
-        print(i)
-    print("\n")
+    print("\n" + "\n".join(list(i for i in donor_dict)))
```
PythonCHB: I haven't talked about this yet, but:
(i for i in donor_dict)

is a "generator exression" -- it's lie a list comprehension, except that it lazy evaluates -- i.e. you need go loop trough it to get the values -- or call list(). but join() will iterate through it too -- so non need for the list.

But in this case, you want the keys, so you can simply do:

"\n",join(donor_dict.keys())

students/gabriel_meringolo/session_04/update_mailroom.py
```diff
@@ -84,129 +83,77 @@ def mail_head(donor_dict):
...
-
-
-def add_donor(x, y):
+def add_donor(donor_dict, donor_name):
     '''
     adds donor to list
-    :param x: donor database list
-    :param y: donor name
-    :return: donor name y added at end of list x
+    :param donor_dict: donor database list
```
PythonCHB: personally, I don't like names like something_dict -- I now it's s dict. but when you document it, you should use "dict", not "list".

students/gabriel_meringolo/session_04/update_mailroom.py
```diff
@@ -84,129 +83,77 @@ def mail_head(donor_dict):
...
-    donation = y[-1]
-    print("Thank you Mr(s). {} for your most genourous donation of ${}".format(name, donation))
-    mail_menu(x)
+    print("Thank you Mr(s). {} for your most generous donation of ${}".format(donor_name, donor_dict.get(donor_name)[-1]))
+    mail_menu(donor_dict)
 
 
 
-def num_check(x, y, z):
+def num_check(donor_dict, donor_name, donation):
```
PythonCHB: either you should refactor, or rename -- if this is checking s number, why does it need the name, and donor dict? a function should generally do just one thing.

students/gabriel_meringolo/session_04/update_mailroom.py
```diff
@@ -84,129 +83,77 @@ def mail_head(donor_dict):
...
     '''
-    z = input("\nEnter donation amount\n> ")
-    if num_check(x, y, z) == True:
-        for i in x:
-            if i[0] == y:
-                i.append(int(z))
-                return i
+    donation = input("\nEnter donation amount\n> ")
+    if num_check(donor_dict, donor_name, donation):
+        (donor_dict.get(donor_name)).append(int(donation))
```
PythonCHB: this kinds misses the point of IBAFTP -- here, you want to convert the input to a number (I"d actually use a float, but whatever...). SO jsut try to do that:

try:
    donation = int(donation)
except ValueError:
    print("you must input a valid number")
    return None # or get back to the prompt somehow
donor_dict[donor_name].append(donation)
....

students/gabriel_meringolo/session_05/dict_list_lab2.py
```diff
@@ -5,14 +5,52 @@
...
 
 
-#Using a list comprehension, build a dictionary of numbers from zero to fifteen and the hexadecimal
-#equivalent (string is fine). (the hex() function gives you the hexidecimal representation of a number.)
+# Using a list comprehension, build a dictionary of numbers from zero to fifteen and the hexadecimal
+# equivalent (string is fine). (the hex() function gives you the hexidecimal representation of a number.)
 
 numbs = [x for x in range(16)]
 hex_numbs = [hex(x) for x in numbs]
 print({x: y for x, y in zip(numbs, hex_numbs)})
```
PythonCHB: why not do this all at once?

``{ x: hex(x) for x in range(16)}``

students/gabriel_meringolo/session_05/dict_list_lab2.py
```diff
@@ -5,14 +5,52 @@
...
+sets = []
+for num in range(2, 5):
+    sets = sets + ["set {}".format(str(num)), {i for i in range(21) if i % num == 0}]
+
+print("Sets: ", sets)
+
+
+# Extra credit: do it all as a one-liner by nesting a set comprehension inside a list
+# comprehension. (OK, that may be getting carried away!)
+# Nope.
```
PythonCHB: :-)

students/gabriel_meringolo/session_05/test_codebat.py
```diff
@@ -0,0 +1,33 @@
+from .codebat import sleep_in, cigar_party
```
PythonCHB: nice tests!

students/gabriel_meringolo/session_06/keyword_arg_lab.py
```diff
@@ -0,0 +1,34 @@
...
+color_code2()
+
+
+def color_code3(fore, back, link="pink", visited="salmon"):
+    print("the fore is {}, the back is {}, the link is {} and the visited is {}.".format(fore, back, link, visited))
+hues = ("red", "rose")
+
+huez = {"link":"Pink","visited": "Salmon"}
+
+color_code3(*hues, **huez)
```
PythonCHB: A little hard to tell what you tried -- but it looks like you've got it.

students/gabriel_meringolo/session_06/lambda_keyword.py
```diff
@@ -0,0 +1,19 @@
+# Write a function that returns a list of n functions,
+# such that each one, when called, will return the input
+# value, incremented by an increasing number.
+# Use a for loop, lambda, and a keyword argument
+
+
+def function_builder(func_amount, l=None):
+    if l is None:
```
PythonCHB: nice touch!

students/gabriel_meringolo/session_06/lambda_keyword.py
```diff
@@ -0,0 +1,19 @@
...
+        l = []
+    for index in range(func_amount):
+        l.append(lambda x, e=index: x + e)
```
PythonCHB: bingo!

PR #96, opened 2015-11-13
Having trouble with part 2 of the trapezoidal lab

elaewin:
In trapezoidal2.py: As it says in my comments, I'm having trouble figuring out how to take arbitrary arguments that are passed into my trapz function and pass those arguments into an enclosed (is that the right term?) version of the quadratic function. Am I trying to go about this the wrong way? If so, I'm at a loss as to how to proceed.


students/erica_winberry/session06/trapezoidal2.py
```diff
@@ -0,0 +1,59 @@
...
+Update your trapz() function so that you can give it a function that
+takes arbitrary extra arguments, either positional or keyword, after the x.
+'''
+
+
+def trapz(fun, start, stop, *args, **kwargs):
+    # compute the area under an arbitrary function,
+    # using the trapezoidal rule.
+
+    if kwargs is not None:
```
PythonCHB: You don't need to check for this -- kwargs will always be a dict -- if just may be empty.

PythonCHB: And you don't want to define the quadratic function in here. It will get passed in as "fun". 

But you do want to pass on args and kwargs to fun here.

The idea is that your trapz2 function can handle ANY function that gets passed in, even if it takes extra arguments.

codedragon:
Is this still confusing or just not had time? Do you want me to go ahead and merge this, so new work will be apparent?

elaewin:
I was out of town all weekend, and didn't have a chance to work on it
much.  Go ahead and merge it, and I'll see if I can't figure it out today.

elaewin:
I'm still not sure how to assign args, kwargs to fun. I tried fun =
fun(*args, **kwargs), but I get an error telling me that quadratic is
missing the positional argument x. It seems to work if I don't put in any
additional arguments (using the values in the quadratic function). Is there
something mathematical that I'm missing? I understand that x is a parameter
in the quadratic function, but not why it's not consistent (that is, it
seems to calculate when I don't pass parameters into trapz() ).

PythonCHB:
So close!

>  I tried fun = fun(*args, **kwargs),
you don't want to re-assign fun -- you want to call it -- WITH the *args, **kwargs

> but I get an error telling me that quadratic is missing the positional argument x. 
yup -- it needs that. it neds x, and A,B,C are optional

It seems to work if I don't put in any additional arguments (using the values in the quadratic function). 


there something mathematical that I'm missing? I think something coding-wise -- this is tricky stuff!

> I understand that x is a parameter in the quadratic function, but not why it's not consistent (that is, it seems to calculate when I don't pass parameters into trapz() ).


OK -- a lot of notes here!

    def quadratic(x, **kwargs):
        return (new_kwargs[a] * x**2) + (new_kwargs[b] * x) + new_kwargs[c]

1) you don't want to define this inside the trapz function -- this is just an example -- trapz should take any function that returns a value and takes at last one positional parameter.

2) when you do: def quadratic(x, **kwargs) -- you are creating a function that requires x, and optional takes any old keyword arguments. -- you don't really want that here -- you want A, B, and C:

    def quadratic(x, A, B, C):
        return (A * x**2) + (B * x) + C

and this way, they are required, so it wont run if you don't call it right :-)
But how to call it right? well, if you want to be able to handle arbitrary arguments, you need:

*args, **kwargs

so:
def trapz(fun, a, b, *args, **kwargs):

now you have to pass in fun, a, and b, and then you can pass in any old thing after those (or ntohgin extra)

in this case the extra stuff wants to be passed through to fun -- so when you call fun:

fun(x, *args, **kwargs)

they'll just be passed on through.

PR #98, opened 2015-11-15
Session Six files

tycobass:
Provides session six files including:
fun_with_functions.py - determines area under a curve
test_fun_with_functions.py
lambda_keyword.py

A whole lot of other stuff including folder named "debris" which includes little snippets of code I generate to help drive home example from the class


codedragon:
This was a little overwhelming. Lambda stuff looks good. I was confused by the fun_with_functions stuff, which seemed to include trapz stuff, that may or may not be the same as the trapz.py.
At any rate, your trapz is not set up to be tested with the tests Chris provided. The idea is to create trapz so it works with the tests, this way you know if you are getting the correct answer, and it makes it easier for us to figure out where you might be going wrong, if you aren't getting your tests to pass.

PR #99, opened 2015-11-16
Files from trapz code

mdpitts2000:
tranz started but still working through tests.


students/martypitts/session06/fodder.dir/trapz.py
```diff
@@ -0,0 +1,253 @@
...
+    :param b: the end point for the integration
+    :type b: a numeric value
+    '''
+    N = list(range(Steps))
+
+    #N = list(range(a,b,StepSize))
+    DeltaX = (b - a) / Steps
+    Xpoint = a
+    TrapCumulativeArea = 0
+    if fun == "sqrt(n-1)":
```
codedragon: You should be able to use the function itself, rather than redefining the function in if statements. Yes, you can pass a function into another function, and then call that passed in function. In other words, the fun argument should not be just a string, but an actual function, and you should call it.

codedragon:
This is a lot of stuff. I looked through it, but it was unclear what you may want help with. I'm going to merge, and then if you can ask questions about ones that you are stuck on, that would be very helpful.

PR #100, opened 2015-11-16
work review trapz.py

ryanmorin:
First part of the homework for reveiw


students/ryan_morin/session_05/trapz.py
```diff
@@ -0,0 +1,30 @@
+__author__ = 'ryan.morin'
+
+import math
+
+def sine(x):
```
PythonCHB: not much point in this, is there :-)

students/ryan_morin/session_06/trapz.py
```diff
@@ -0,0 +1,30 @@
...
+def sine(x):
+    return math.sin(x)
+
+def line(x):
+    return 5
+
+def quadratic(x, c=0, d=0, e=0):
+    return c * x**2 + d*x + e
+
+def trapz(fun, a, b, *args, **kwargs):
```
PythonCHB: so you've got the extra arguments -- but you need to do something with them ...

students/ryan_morin/session_06/trapz.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    total = 0.0
+    inc = float((b-a))/100
+    total += fun(a)/2.0
+    for i in range(1,100):
+        total += fun(a + (i * inc))
+    total += (fun(b)/2.0)
+    return total * inc
+
+print (trapz(lambda x: math.sin(x), 1,10))
```
PythonCHB: math.sin is already a function, so no need to wrap it in a lambda.

PythonCHB:
nice work!

PR #101, opened 2015-11-17
trapz stage 2 with tests

robalford:


students/robalford/session06/test_trapz.py
```diff
@@ -55,104 +56,140 @@ def test_is_close():
...
-def test_trapz():
+def test_line():
     # a simple line
     assert trapz(line, 0, 10) == 50
+
+
+def test_curves():
     # a simple curve
     under_curve = trapz(a_curve, 1, 6)
     assert isclose(under_curve, 7.4512822710374)
```
PythonCHB: how did you get these values to test against? Ideally you have another way to confirm the correct answer -- though that's not always possible.

students/robalford/session06/trapz.py
```diff
@@ -19,64 +19,42 @@ def another_curve(x):
 def quadratic(x, A=0, B=0, C=0):
     return A * x**2 + B * x + C
 
-coef = {'A': 1, 'B': 3, 'C': 2}
-
 
 def trapz(fun, a, b, **coef):
```
PythonCHB: if you want to be able to pass positional arguments in as well -- you need *args and **kwargs.

and I'd use **kwargs, rather than **coef for the name -- in the quadratic case, they are coefficients, so good name -- but in the arbitrary case, they could be anything. And the trapz function does know what they are -- only that it will pass them along.

students/robalford/session06/trapz.py
```diff
@@ -19,64 +19,42 @@ def another_curve(x):
...
     num_of_grids = 100
     grid_spacing = (b-a)/num_of_grids
-    # return grid_spacing
     list_of_values = [a + grid_spacing]
     for i in range(num_of_grids-2):
         list_of_values.append(list_of_values[-1] + grid_spacing)
```
PythonCHB: this is risky -- by sequentially adding the grid_spacing, you are accumulating any error in the value -- it can add up.

students/robalford/session06/trapz.py
```diff
@@ -19,64 +19,42 @@ def another_curve(x):
...
-#     grid_spacing = (b-a)/num_of_grids
-#     # return grid_spacing
-#     list_of_values = [a + grid_spacing]
-#     for i in range(num_of_grids-2):
-#         list_of_values.append(list_of_values[-1] + grid_spacing)
-#     # return list_of_values
-#     list_of_values = [2*fun(i) for i in list_of_values]
-#     list_of_values.append(fun(a)+fun(b))
-#     return (b-a)/(2*num_of_grids) * sum(list_of_values)
+def dynamic_trapz(fun, a, b, num_of_grids=100):
```
PythonCHB: this lets the user pass in the resolution the want -- nice feature.

But what I was going for was having the function decide for itself how many to use.

See my solution

PR #103, opened 2015-11-18
Still working on args, kwargs issues

elaewin:


PythonCHB:
I'll commit this, so we can look at it in class -- thanks!

PR #104, opened 2015-11-22
Session 7 HW

robalford:


students/robalford/session07/html_render.py
```diff
@@ -0,0 +1,123 @@
...
+            self.content.append(content)
+
+    def append(self, content):
+        self.content.append(content)
+
+    def render(self, f, ind='    '):
+        f.write(ind * self.indent)
+        start_tag = '<{}'.format(self.tag)
+        # is there a better way to format this?
+        # also should inherit this portion of render via
```
PythonCHB: Good catch -- yes, factoring this out to be used in other render methods is a great idea.

You can do that by making a new method: format_tag() or something. Then call that from the all the render methods.

students/robalford/session07/html_render.py
```diff
@@ -0,0 +1,123 @@
...
+        # another method? rather than copy
+        for k, v in self.attributes.items():
+            start_tag += ' {}="{}"'.format(k, v)
+        start_tag += '>'
+        f.write(start_tag)
+        for el in self.content:
+            f.write('\n')
+            # EAFP
+            try:
+                el.indent = self.indent + 1
```
PythonCHB: Careful here -- this could get out of sync easily-- why not pass the indent level through the render method. It's not state that needs to be preserved.

robalford: I think I might be getting tripped up by the instructions for the homework here, that "the amount of indentation should be set by the class attribute: indent". In my code, I'm trying to ensure that each nested tag will always be indented one level further than it's containing tag. These indent values can't be fixed because, for example, every p tag will not always be indented to the same level in an html doc. That's why I'm confused by setting the indent level as a class attribute, rather than an instance variable that might be a different value for each instance of the tag. 

In my code, I start with an indentation level of 0 for the DOCTYPE in Html class, then increment it to 1 for the html tag. In the render method, I multiplying the ind argument (a single tab space) by the class attribute indent. Then, I want every nested tag that is stored in the content list to have an equal indent level 1 greater than it's containing tag. That's why I increment the indent level here, within the for loop that is acting on the nested tags, so that when it returns to the beginning of the render method, it will multiply the tab space by an increment of one, and continue to do so for deeper nested elements. 

I'm sure I'm missing an obvious simpler solution here. My code does seem to be working with the test html, but I see it's problematic to be storing these dynamic values in the class attributes, and doing it this way could lead to trouble in a more complex document. Maybe I need a clearer understanding of class vs. instance attributes, or maybe I just misunderstood the instructions for the homework.

cewing: Can you think of a way to accumulate indentation as you work your way down into a set of nested tags?  Regardless of what the tag type is, you are correct in saying that the indentation level  should be based on context.  So how can you pass that context around?

PythonCHB: On Mon, Nov 23, 2015 at 10:29 AM, robalford <notifications@github.com>
wrote:

>  "the amount of indentation should be set by the class attribute: indent".
> In my code, I'm trying to ensure that each nested tag will always be
> indented one level further than it's containing tag. These indent values
> can't be fixed because, for example, every p tag will not always be
> indented to the same
>
level in an html doc.
>
yup -- sorry that wan't clear -- what I meant was that the amount of
indentation *per level* be set as a class attribute -- i.e. two spaces per
level, four spaces per level, etc...

> That's why I'm confused by setting the indent level as a class attribute,
> rather than an instance variable that might be a different value for each
> instance of the tag.
>
ah, but you don't want each instance to have a different level of
indentation -- that's really a render-time issue. conceivable, if you had
some repeated tags, you might use the same instance in more than one place
in the docs -- or more than one doc. So the level of indentation is a realy
a render-time value.

See Cris' note for a hint....

-CHB


-- 
Christopher Barker, PhD

Python Language Consulting
  - Teaching
  - Scientific Software Development
  - Desktop GUI and Web Development
  - wxPython, numpy, scipy, Cython

students/robalford/session07/html_render.py
```diff
@@ -0,0 +1,117 @@
...
+    def append(self, content):
+        self.content.append(content)
+
+    def format_start_tag(self, self_closing=False):
+        start_tag = '<{}'.format(self.tag)
+        for k, v in self.attributes.items():
+            start_tag += ' {}="{}"'.format(k, v)
+        start_tag += ' />' if self_closing else '>'
+        return start_tag
+
```
codedragon: Nice solution for formatting the start tag.

students/robalford/session07/test_html_render.py
```diff
@@ -0,0 +1,247 @@
...
+    text = f.read().strip()
+    assert text.startswith('<html>')
+    assert text.endswith('</html>')
+    # these tests need work. adjust indent of <p> has no affect on test passing
+    assert '    <html>' not in text
+    assert '    <body>\n' in text
+    assert '    </body>' in text
+    assert '        <p>' in text
+    assert '        </p>' in text
+    # print(text)
```
codedragon: Assert ... in will only catch if you have at least that amount of space before <html>, but won't check for that exact amount of space. Instead try splitting the text on \n, so that you get lines of text separated, and then check individual lines for starting space with startswith.

codedragon:
Looks good. Looks like you have a good understanding of when and how to override parent class methods.

PR #106, opened 2015-11-24
html render

gabrielx52:
html render works. I still need to finish the pretty printing and I'm going to refactor it a bit. I feel like the multiple render methods can be reduced.


students/gabriel_meringolo/session_07/html_render.py
```diff
@@ -0,0 +1,125 @@
...
+        self.attributes = kwargs
+        self.content = []
+        if content is not None:
+            self.content.append(content)
+
+    def append(self, content):
+        self.content.append(content)
+
+    def render(self, f, ind="    "):
+        if self.attributes != {}:
```
codedragon: if not self.attributes:  will do the same thing as  if self.attributes != {}:

codedragon:
I suspect you won't be able to reduce the number of render methods, as it looks like you have added them to the classes that need them, but you should be able to have fewer lines of code in some of them. Think about your loops, and when/if you want to render vs. write. You are getting the hang of it, good work.

PR #108, opened 2015-11-24
Homework, Week 7

elaewin:
Not much to ask about. I got the renderer to work, for the most part.  Looking at the solution, my code looks very, very cluttered.

### DIFF
--- a/students/erica_winberry/session07/html_render.py
+++ b/students/erica_winberry/session07/html_render.py
@@ -1,19 +1,27 @@
 class Element:
 
     indent = 0
-    tag = "html"
+    tag = ""
 
-    def __init__(self, content=None):
+    def __init__(self, content=None, **kwargs):
         self.content = []
         if content is not None:
             self.content.append(content)
+        self.kwargs = kwargs
 
     def append(self, content):
         self.content.append(content)
 
     def render(self, f, ind=" "):
-        start_tag = "\n<{}>".format(self.tag)
-        f.write(start_tag)
+        start_tag = "\n<{}".format(self.tag)
+        if self.kwargs:
+            f.write(start_tag)
+            for k, v in self.kwargs.items():
+                attribute = '{}="{}"'.format(k, v)
+                f.write(" " + attribute)
+            f.write(">")
+        else:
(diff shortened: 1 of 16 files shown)


students/erica_winberry/session07/html_render.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class Link(Element):
+    """Renders link (anchors) element."""
+
+    indent = 0
+    tag = "a"
+
+    def __init__(self, link=None, content=None, **kwargs):
+        Element.__init__(self, content=None)
```
PythonCHB: you are calling the Element __init__ here -- good. but then that means you don't need to re-produce what it does here.

i.e. everything but setting the link should be done.

students/erica_winberry/session07/html_render.py
```diff
@@ -0,0 +1,242 @@
...
+    tag = "ul"
+
+
+class OneLineTag(Element):
+    """For tags that should always render on a single line."""
+
+    def render(self, f, ind=" "):
+        start_tag = ("\n{}<{}".format((ind * self.indent), self.tag))
+        if self.kwargs:
+            f.write(start_tag)
```
PythonCHB: haven't I seen this code somewhere before :-) remember DRY: can you factor out this attribute writing code to a shared method?

students/erica_winberry/session07/test_html_output8.html
```diff
@@ -0,0 +1,26 @@
...
+        <ul style="line-height:200%" id="TheList">
+
+            <li>
+                The first item in a list
+            </li>
+            <li style="color: red">
+                This is the second item
+            </li>
+            <li>
+                And this is a <a href="http://google.com">link</a>                to google
```
PythonCHB: intereting -- looks like some extra space in there -- an indentation that we don't want?

PythonCHB:
nice work. get it to work, then go in and see if you can clean it up -- biggest thing is DRY -- if you have the same code repeated in multiple places, try to refactor that out to a separate function.

Personally, its when I copy and paste it the second time -- In know I should refactor :-)

PR #112, opened 2015-11-27
sparse array 

gabrielx52:
I've added the len() fun to the SparseArray class but I'm conflicted on the direction I should proceed from here...
 The idea behind the exercise is to create a special class thats helps conserve memory when dealing with sparse arrays by removing the zeros. Would keeping the original sparse array around so it can be used for the len() fun defeat the purpose of the lab? or would creating a len var that increases every time a number is added ;regardless if a zero or not; be a better route?

### DIFF
--- a/students/gabriel_meringolo/session_08/sparse_array.py
+++ b/students/gabriel_meringolo/session_08/sparse_array.py
@@ -3,24 +3,43 @@
 class SparseArray:
 
     def __init__(self, *args):
-        self.sparse = args
         self.array = {}
-        self.length = len(args)
-        for i in self.sparse:
+        self.length = 0
+        self.sparse = []
+        for i in args:
+            self.length += 1
             if i != 0:
-                self.array[self.sparse.index(i)] = i
-
-    def __repr__(self):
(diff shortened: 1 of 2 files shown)


PythonCHB:
>Would keeping the original sparse array around so it can be used for the len() fun defeat the purpose of the lab?

I'm afraid so -- you'll need to keep a counter to keep track of how big the "virtual" array is.

CHB

students/gabriel_meringolo/session_08/sparse_array.py
```diff
@@ -0,0 +1,45 @@
...
+class SparseArray:
+
+    def __init__(self, *args):
+        self.array = {}
+        self.length = 0
+        self.sparse = []
+        for i in args:
+            self.length += 1
+            if i != 0:
+                self.array[args.index(i)] = i
```
codedragon: Alternatively, you could do:
for ind, data in enumerate(args):
    self.length += 1
    if data != 0:
        self.array(ind) = data

codedragon:
Looks good!

PR #113, opened 2015-11-29
In class labs

ryanmorin:
The in class labs.  I'm still working on the circle lab.

### DIFF
--- a/students/ryan_morin/session_08/Circle.py
+++ b/students/ryan_morin/session_08/Circle.py
@@ -1,0 +1,48 @@
+__author__ = 'ryan.morin'
+
+
+"""
+ A nifty Circle class
+ """
+
+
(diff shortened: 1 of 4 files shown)


students/ryan_morin/session_08/Circle.py
```diff
@@ -0,0 +1,48 @@
...
+     def __str__(self):
+         return 'Circle with radius: {}'.format(self.radius)
+
+     def __repr__(self):
+         return 'Circle({})'.format(self.radius)
+
+     def __add__(self, other):
+         total = self.radius + other.radius
+         return total
+
```
codedragon: I think the idea is to return a circle with the new radius, rather than just the new radius.

codedragon:
Off to a good start.

PR #114, opened 2015-11-29
Step 6 of circle lab question

elaewin:
(In circle_lab.py and test_circle.py) I've written __str__ and __repr__ methods, and they work properly, but my test fails. Not sure what I'm doing wrong. Am I missing something?

### DIFF
--- a/students/erica_winberry/session08/circle_lab.py
+++ b/students/erica_winberry/session08/circle_lab.py
@@ -28,3 +28,14 @@ class Circle:
 
     def __repr__(self):
         return "Circle({:.0f})".format(self.radius)
+
+    def __add__(self, other):
+        new_radius = self.radius + other.radius
+        return Circle(new_radius)
+
+    def __mul__(self, other):
+        new_radius = self.radius * other
+        return Circle(new_radius)
+
+    def __rmul__(self, other):
+        return self.__mul__(other)
--- a/students/erica_winberry/session08/circle_lab.py
+++ b/students/erica_winberry/session08/circle_lab.py
@@ -33,9 +33,22 @@ class Circle:
         new_radius = self.radius + other.radius
         return Circle(new_radius)
 
+    def __sub__(self, other):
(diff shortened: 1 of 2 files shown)


students/erica_winberry/session08/test_circle.py
```diff
@@ -24,16 +24,61 @@ def test_diameter2():
...
+
+def test_from_diameter():
+    c = cl.Circle.from_diameter(8)
+    assert c.diameter == 8
+    assert c.radius == 4
+
+
+def test_str():
+    c = cl.Circle(4)
+    text = print(c)
```
PythonCHB: print() will print to the screen, but it doesn't return the string printed -- it returns None.

so you want:

text = str(c)

print() calls str() under the hood.

students/erica_winberry/session08/test_circle.py
```diff
@@ -24,16 +24,61 @@ def test_diameter2():
...
+    assert "Circle with radius: 4.000000" in text
+
+
+def test_repr():
+    c = cl.Circle(4)
+    text = repr(c)
+    assert "Circle(4)" in text
```
PythonCHB: this looks like it should work -- does it?

students/erica_winberry/session08/circle_lab.py
```diff
@@ -18,3 +18,37 @@ def diameter(self, value):
...
+
+    @classmethod
+    def from_diameter(cls, value):
+        return cls(value / 2)
+
+    def __str__(self):
+        return "Circle with radius: {:.6f}".format(self.radius)
+
+    def __repr__(self):
+        return "Circle({:.0f})".format(self.radius)
```
PythonCHB: {:.0f} means: format a floating point number with 0 decimal places -- probably not what you want.

The idea of __repr__ is to be abel to re-create the object -- so you want a lot of digits. Maybe just:

{:f}

PR #115, opened 2015-11-30
Added circle lab

iampei:


students/pei/session08/circle.py
```diff
@@ -0,0 +1,67 @@
...
+
+# step 8
+    def __gt__(self, c):
+        return self.radius > c.radius
+
+    def __eq__(self, c):
+        return self.radius == c.radius
+
+# step 9
+    #doesn't seem to need to do anything
```
codedragon: Step 8 had the side effect of giving sort the tools it needs to sort, so you are good!

PR #116, opened 2015-12-01
Final project starting code

ericrosko:
This pull request includes the beginning of my final project.  Please check out the readme.md file inside of the session10 folder for the complete description.

I've also fixed the html render assignment according to the pep8 guidelines.

Thanks.


PythonCHB:
whoa! lots of changes -- did you clean up the tab indents?

students/ericrosko/session10-final-project/ReadMe.md
```diff
@@ -0,0 +1,40 @@
...
+A common way this happens is that an assignment has some included assets, like icon files.  The student downloads them to their Mac's downloads folder, then adds them to their project without selecting copy from their file copy dialog box.  Now Xcode will look in /home/student-username/downloads when I open the project on my computer, and of course it can't find the file.  Then I either fix their project myself with a copy of the files I have, or send an email asking them to resubmit.
+
+# Implementation Details
+
+The Xcode project file is just a package, so I can browse into it just like it is possible with a directory.  This makes it possible to parse the part of the project file which contains all file references.  The file is just plain text which makes it easy.
+
+# Goals
+* Find any images that are not included in the project folder
+* Find any images that exist in the project but are not needed by the project
+
```
PythonCHB: great project idea -- manageable and useful!

students/ericrosko/session10-final-project/asset_checker.py
```diff
@@ -0,0 +1,208 @@
...
+__author__ = 'Eric Rosko'
+
+
+class AssetChecker():
+
+    ignored_directories = ['__pycache__', '.cache']
+    manifest_set = {}
+    project_set = {}
+
+    def __init__(self, *args, **kwargs):
```
PythonCHB: what might be in *args? if it's a list of paths, for instance, it might be better to simpy use:

__init__(self, path_list):

It makes it much clearer for the user to know what's expected.

and don't use **kwargs unless you are going to do something with them!

students/ericrosko/session10-final-project/asset_checker.py
```diff
@@ -0,0 +1,208 @@
...
+        self.starting_search_path = os.getcwd()
+        self.searchable_extensions = ['m4a', 'jpg', 'png']
+
+        for item in args:
+            if isinstance(item, str):
+                self.starting_search_path = item
+            elif isinstance(item, list):
+                self.searchable_extensions = item
```
PythonCHB: OK -- kind of a klunky API -- first, I may not be right about what's going on here, which shows that it's a hard-to understand API, but if I am right, then I'd do:

def __init__(self, search_path, searchable_extensions):

type switching on arbitrary arguments is pretty fragile -- what if the user passes in a tuple of searchable extensions? Your code will fail, but it's really better to be flexible with that.

students/ericrosko/session10-final-project/asset_checker.py
```diff
@@ -0,0 +1,208 @@
...
+            else:
+                raise Exception("Bad Parameter:{}".format(item))
+
+    def search_tree(self, path, files, extensions):
+        assert isinstance(extensions, list)
```
PythonCHB: again -- I know you like type safety -- but don't do this -- really!

If you really really want this (I say don't bother), then you want to check against an Abstact Base Class:

isinstance(l, extensions.Sequence)

https://docs.python.org/3/library/collections.abc.html

Better yet, probably:

 isinstance(l, collections.Iterable)

if all you need to do is be able to iterate over it. (i.e. not add to it, etc.)

cewing: It's also worth keeping in mind when using `assert` in run-time code that there is a way of executing Python that removes all `asserts`: the `-O` or `-OO` flag.  See [this blog post](http://blog.ziade.org/2015/11/25/should-i-use-pythonoptimize/) for some thoughts about it.

students/ericrosko/session10-final-project/asset_checker.py
```diff
@@ -0,0 +1,208 @@
...
+                results.extend(found_items)
+                # print("length: ", len(results))
+
+    def compiled_regexes(self):
+        assert len(self.searchable_extensions) >= 1, "Expected at least " \
+            "one searchable extension (jpg, etc)"
+        regexes = []
+
+        for ext in self.searchable_extensions:
+            regexes.append(re.compile("[A-Za-z0-9_-]+.{}".format(ext)))
```
PythonCHB: do you need a regex for this? could you simpily do: something.endswith("."+ext)

or something like that? regexs have their place, but for the simple stuff, you can often just use python's string methods.

though if you think you may need to get fancier, having a bunch of pre-complied regex's is a good way to go.

students/ericrosko/session10-final-project/asset_checker.py
```diff
@@ -0,0 +1,208 @@
...
+
+        return regexes
+
+    def perform_asset_audit(self):
+        temp = []
+        self.find_project_file(self.starting_search_path, temp)
```
PythonCHB: if self already has a starting_search_path attribute,t hen you shouldn't need to pass it in. KInd of the point of OO development.

students/ericrosko/session10-final-project/asset_checker.py
```diff
@@ -0,0 +1,208 @@
...
+
+        for item in args:
+            if isinstance(item, str):
+                self.starting_search_path = item
+            elif isinstance(item, list):
+                self.searchable_extensions = item
+            else:
+                raise Exception("Bad Parameter:{}".format(item))
+
+    def search_tree(self, path, files, extensions):
```
PythonCHB: take a look at os.path.walk() -- could be helpful here.

or various version on the internet for walking a pathlib path, for isntance:

https://gist.github.com/abingham/47a9382f42d6cb3643eb

students/ericrosko/session10-final-project/asset_checker.py
```diff
@@ -0,0 +1,208 @@
...
+                    if "/* End PBXBuildFile section */" == line:
+                        doParse = False
+                    elif "/* Begin PBXBuildFile section */" == line:
+                        doParse = True
+                    # print(line)
+                    if doParse:
+                        self.parse_line(line, regexes, results)
+                    # print(results)
+
+        except Exception:
```
PythonCHB: avoid catching a generic Exception! you'll lose the information about what when wrong.

In this case, I"d probably not catch it at all -- the whole script will abort, but with useful inforation.

or you could catch particular Exceptions if you want to make a nice message. There are  bunch for file IO issues, like:

FileExistsError
FileNotFoundError

Oo, at least, catlch Exception, write a nice mesage, then re-raise it:

except:
    print("file read failure")
   raise

PythonCHB:
merging what you have so far.

students/ericrosko/session10-final-project/asset_checker.py
ericrosko: I wasn't sure if I would use them, and even thought I had to have the init signature look like that.  It appears I can delete kwargs** and the script still works.

students/ericrosko/session10-final-project/asset_checker.py
ericrosko: Sounds good I'll add those as named parameters.  If the user passes in the wrong thing in the the program will crash and they'll know they did something wrong. ;-).  I do have some error checking in my __main__ function at the bottom of the script now, but yes it should be bulletproof.

students/ericrosko/session10-final-project/asset_checker.py
ericrosko: I wanted the user to be able to enter in a path or list of arguments in whatever order they wanted, so the idea was just looping through everything and just picking them out as it loops.

I added the is instance assert in this case to make sure I was passing in a list.  I figure if I needed speed I could also optimize it away, and if I leave it in there it goes unnoticed.  I like to use asserts to keep myself from basic mistakes, and then it behaves like a print statement that only prints if something goes wrong.  I would rather use an assert than add a print statement in the code that I comment out later or leave in.  I've had some great success with this, and it gives me some peace of mind, though I realize it could drive some people crazy.</holywar>

I wouldn't ever use an assert to direct program flow though.  That's a great article.  I'd would always optimize out the asserts  and run it just to make sure there wasn't a side effect of the assert which breaks it.

The comments in that article mention if __debug__: log.debug("blah %s", arg).  That might be good for me to check out.

I just ran my script with python3 -OO and it still ran.  I'm not sure where the .pyc or .pyo files are supposed to go, they aren't in my folder or the __pycache__.

students/ericrosko/session10-final-project/asset_checker.py
ericrosko: I liked how the regex returned an array of every match it found in the line, so that was the winning feature for me.  I suppose I could have split it by space and then used 'endswith' as a loop through every string in the returned array unless you know of a better way.  I had to either precompile the regex functions, or do it every since time for every loop, so that's why I have a function to just do it once.

students/ericrosko/session10-final-project/asset_checker.py
ericrosko: Yeah, true.  I wrote the constructor last so that is my excuse. :-)

students/ericrosko/session10-final-project/asset_checker.py
ericrosko: Good point.  I know better than to swallow exceptions silently.  Here is what I did above:
        except Exception as e:
            raise e

In your code snipped you just do:
except:
print("file read failure")
raise

So I guess you don't need to reference it with 'e'?

PR #117, opened 2015-12-02
Adding additional files that were not previously managed

maxmac12:
Please only review my final project propsal, my Circle class and the associated test file, and my initial HTML renderer file and associated test file. I'm not sure how to remove certain files from a pull request since most of the others I don't want to include.


PythonCHB:
thanks -- you're right, you can't make a pull request for only certain files -- it does a diff on teh whole repo.

Actually, you can kind of do that by putting certain files in a branch -- but that's more pain than it's worth. We are using this in a kind of weird way -- so just adding a note is fine.

students/maxmaccamy/Final Project/Final Project Proposal.txt
```diff
@@ -0,0 +1,5 @@
+For my final project in For my final project in this class, I would like to make a program that traces low level requirement tags used for my job
+to the location of those tags within our code base. The input to the program will be a CSV type input that contains all of
+the requirement tags that should be traced somewhere into our code base. The outputs will be a report document that shows
+in which file and at which line number the trace tag is located, as well as an error report indicating which trace tags are
+missing so that the user can go in and add the missing tags a at some date.this class, I would like to make a program that traces low level requirement tags used for my job to the location of those tags within our code base. The input to the program will be a CSV type input that contains all of the requirement tags that should be traced somewhere into our code base. The outputs will be a report document that shows in which file and at which line number the trace tag is located, as well as an error report indicating which trace tags are missing so that the user can go in and add the missing tags a at some date.
```
PythonCHB: This looks like a great idea for a project - modest in scope but useful.

students/maxmaccamy/session07/html_render.py
```diff
@@ -0,0 +1,46 @@
+__author__ = 'Max'
```
PythonCHB: this all looks good!

students/maxmaccamy/session07/test_html_render.py
```diff
@@ -0,0 +1,105 @@
...
+
+def test_html():
+    e = hr.Html("this")
+    f = StringIO()
+    e.render(f)
+    f.seek(0)
+    text = f.read().strip() # Remove trailing white space
+    assert text.startswith("<html>")
+    assert text.endswith("</html>")
+    assert "this" in text
```
PythonCHB: I know I started doing it this way in class, but once you find yourself writing virtually the same code three times -- it's time to factor that out into a stand alone test function that you can call from each test. I di da bitof this in my solutonl though you could probably do even more than I did.

students/maxmaccamy/session08/circle.py
```diff
@@ -0,0 +1,50 @@
...
+    def diameter(self, value):
+        self.radius = value / 2
+
+    @property
+    def area(self):
+        return (math.pi * (self.radius ** 2))
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        return Circle(diameter/2)
```
PythonCHB: Here you want to return  ``cls(diameter/2)`` -- in this case, it will be the same thing, but ``cls`` is the actual class object at the time this is run -- which could be a subclass of Circle, rather than Circle itself -- so using it allows this class method to work for subclasses as well.

students/maxmaccamy/session08/circle.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __str__(self):
+        return "Circle with radius: {}".format(self.radius)
+
+    def __repr__(self):
+        return "Circle({})".format(self.radius)
```
PythonCHB: one thing to think abou there is how many digits of accuracy you want to show -- one way to make that easy is to use repr() in your __repr__:

"Circle({})".format(repr(self.radius))

that way you'll get the "proper" accuracy that the number type uses.

(I think there's a way to get the repr version in a format string too -- but I"d have to look that up.

students/maxmaccamy/session08/circle.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __eq__(self, other):
+        return (self.radius == other.radius)
+
+    def __gt__(self, other):
+        return (self.radius > other.radius)
+
+    def __lt__(self, other):
+        return (self.radius < other.radius)
+
```
PythonCHB: this all looks good.

students/maxmaccamy/session08/test_circle.py
```diff
@@ -0,0 +1,80 @@
...
+    assert not (c1 == c3)
+
+def test_gt():
+    c1 = Circle(1)
+    c2 = Circle(1)
+    c3 = Circle(2)
+    assert not (c1 > c2)
+    assert (c3 > c1)
+
+def areaOfCircle(r):
```
PythonCHB: nice use of utility function

students/maxmaccamy/session08/test_circle.py
```diff
@@ -0,0 +1,80 @@
...
+def test_add():
+    c1 = Circle(2)
+    c2 = Circle(4)
+    newCircle = c1 + c2
+    assert (6 == newCircle.radius)
+    assert (12 == newCircle.diameter)
+
+def test_mul():
+    c = Circle(2)
+    c *= 2
```
PythonCHB: ideally *= is an "in place" opeartor for mutable objects -- and this is mutable. You can define an __iadd__ method, and have it change the circle in place and return itself. then test that with something like:

c = Circle(2)
c *= 2
assert c.radius == 4

students/maxmaccamy/session08/test_circle.py
```diff
@@ -0,0 +1,80 @@
...
+    c3 = Circle(2)
+    assert (c1 == c2)
+    assert not (c1 == c3)
+
+def test_gt():
+    c1 = Circle(1)
+    c2 = Circle(1)
+    c3 = Circle(2)
+    assert not (c1 > c2)
+    assert (c3 > c1)
```
PythonCHB: one thing to try (and test for) is tp put a bunch of circles in a list and call list.sort() -- and make sure it works.

maxmac12:
My html_render.py file is not complete but I appreciate your comments.
Unfortunately I probably won't get back of my previous files based on time
limitations; nonetheless all of your comment have been noted. For
the cls(diameter/2) from circle.py you're comments make sense. Instead of
making a new object just operate on the one you were passed. I need to
remember that every variable ("name") is static until the program is
killed. I'm blaming my c/c++ background where I need to manage garbage
keeping manually.

For the your comments on test_circye.py "ideally *= is an "in
place" operator for mutable objects -- and this is mutable. You can define
an *iadd* method, and have it change the circle in place and return itself.
then test that with something like:

c = Circle(2)
c *= 2
assert c.radius == 4

What was I doing wrong or simply an improvement on your part?

Thanks for all your help,


Max

PythonCHB:
Not strictly wrong -- as you saw, it worked fine. You can overload operators to do whatever makes sense for your use case. But it's the convention that iadd is usually a mutating operation.

PR #118, opened 2015-12-02
Project Synopsis

tycobass:
In my household we have multiple computers all backed up to several location with several copies. 
My project will be a build a file crawling tool that will discover duplicate files in a given file 
structure and list them for potential deletion.  Eventually, I want to do the same thing with pictures 
on my storage devices at home.
The project will be developed in the students/eric_v/Session_08/Project file.


PythonCHB:
Sounds like a great project. Have you yet figured out how to keep the data of what files go where?

I've done a simlar thing for uploading photos from my digital camera(s). when I upload them, I store a hash of each photo in a big dict. Then the next time I upload, I can check the hash to see if that particular picture has already been uploaded. IF not, it get uploaded, and added to the dict.

when done, I store the dict in a single file as a pickle -- see:

http://pythonchb.github.io/PythonTopics/persistance_serialization.html

for a bit on that.

for hashing, see: https://docs.python.org/3.5/library/hashlib.html

you probably don't need a big, secure hash for this, but there is little cost -- reading the files from disk is more expensive than computing the hash anyway. md5 is probably fone, I but I'm think I'm using sha256 -- cause why not?

The other issue is really big files, like video -- in that case, I just ash the first MB -- I figure it's pretty darn unlikely to have to files that are exactly the same in the first MG that aren't the some all the way through -- though you might want to think about if you have video clips that are subsets of longer video files...

I'm interested to see how this turn out -- in fact, I've got  alot of photot in my archive ethat I put in before I had my downloaded, or hat have been put in by hand, and I"ve been meaning to write a de-duplicator for that.... maybe you'll write it for me ;-)









for more on that

PR #120, opened 2015-12-05
Week 9 homework

elaewin:
Finished the labs that were started in class.


students/erica_winberry/session09/generator_lab.py
```diff
@@ -0,0 +1,45 @@
...
+    count = 0
+    while x < 16:
+        yield x
+        count += 1
+        x += count
+
+
+def doubler():
+    x = 1
+    while x < 33:
```
PythonCHB: why 33? this could be "while True" -- or, pass the max in to the function if you don't want it to run on forever....

students/erica_winberry/session09/generator_lab.py
```diff
@@ -0,0 +1,45 @@
...
+    yield n
+    yield oldfib
+    while n < 35:
+        newfib = n + oldfib
+        yield newfib
+        n = oldfib
+        oldfib = newfib
+
+
+def prime():
```
PythonCHB: you should be able to find a way to have this work for any number of primes.... (even it will get real slow)

students/erica_winberry/session09/iterator_2.py
```diff
@@ -0,0 +1,45 @@
...
+"""
+
+
+class IterateMe_2(object):
+    """
+    About as simple an iterator as you can get:
+    returns the sequence of numbers from zero to 4
+    ( like range(4) )
+    """
+    def __init__(self, start=0, stop=10, step=1):
```
PythonCHB: you probably want to require a stop parameter. Kind of tricky, actually , because if only one argument, then it is start, but if two then the fist is start, and the seconds is stop....

students/erica_winberry/session09/iterator_2.py
```diff
@@ -0,0 +1,45 @@
...
+        # self.current = start - step  NOTE: Removed along with iter and next.
+        self.start = start
+        self.stop = stop
+        self.step = step
+
+    def __getitem__(self, position=0):
+        # Makes this like range
+        current = self.start + position * self.step
+        if current > self.stop:
```
PythonCHB: to be like range() you want >=

PythonCHB:
Nice work -- looking forward to seeing how the game works out.

-Chris

PR #122, opened 2015-12-06
Final project first attempt

ryanmorin:
This is the first attempt at the final project. I'm working on a different version using pandas.


PythonCHB:
Looks like a fine start -- I"ll wait 'till you are done to look closer.

PR #123, opened 2015-12-06
Final project

ericrosko:
I fixed the issues raised in your comments from the first pull request when I started the final project.

I'm wondering if there is a better way of dealing with arguments passed into a script, or if the way I have it is fine in the asset_checker.py's if __name__ == '__main__' method.  Your suggesting of naming parameters really simplified the code.  I was under the impression that we should use an (*args, **kwargs) pattern for our initializer, but this looks like its just a basic pattern that isn't enforced or preferred by the language like C's argc, argv[] requirement.

I installed the pep8 and sublimelinter pyflakes plugins in Sublime Text 3 so that should explain while the code looks uniform.  It drove me crazy and first but I'm a believer now.  Your little push to get my tabs into spaces was good, as I went and spent time customizing my own user preferences file and I figured out the package manager so that hurdle is over.

Thanks.


PythonCHB:
Looking good!

For command line apps, there is the standard library argparse module,  but I think this is the way to go:

http://docopt.org

Really nice and easy command line argument system.

PythonCHB:
> or preferred by the language like C's argc, argv[] requirement.

well, that's only for the main() function, yes? 

In python, there is no main() convention -- the code in a module simply gets run as a top level script, so there is nothing to pass argc to. As you discovered, the command line gets put in sys.argv if you want to do something with it.

As for regular classes -- the usual convention is to pass in the parameters the initializer needs.

*args and **kwargs is used for when you need to take arbitrary arguments (or pass arbitrary arguments on to a superclass.

students/ericrosko/session10-final-project/asset_checker.py
```diff
@@ -10,52 +24,65 @@
...
+    # any duplicates since sets cannot have the same element twice
     manifest_set = {}
     project_set = {}
 
-    def __init__(self, *args, **kwargs):
-        self.starting_search_path = os.getcwd()
-        self.searchable_extensions = ['m4a', 'jpg', 'png']
+    def __init__(self, search_path=None, searchable_extensions=None,
+                 show_all_output=None):
+        if search_path is None:
```
PythonCHB: People often use:

self.search_path = os.getcwd() if search_path is None else search_path

Taste really, but it's common enough idiom to that there is a PEP to add a simpler way to spell it.

PythonCHB:
Nice work! love the tests.

PR #126, opened 2015-12-09
final project submission

HasBob:


students/bob/project-np/np.py
```diff
@@ -0,0 +1,94 @@
...
+    ssh = paramiko.SSHClient()
+    ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy()) # allow/auto-add all hosts attempted
+
+    print(
+'''
+Warning: this module automatically adds 
+all listed hosts to the known_hosts file.  
+Only use this in lab or on trusted networks.\n'''  # probably not the best place for this warning to go...
+    )
+
```
codedragon: Yes. If you were making this into a module for distribution, it should go in the README, and probably again printed out when you enter the np function.

students/bob/project-np/np.py
```diff
@@ -0,0 +1,94 @@
...
+            
+            stdin, stdout, stderr = ssh.exec_command(commands)
+            #print(stderr.readlines())
+            for line in stdout:
+                line = stdout.readline()
+                if i not in output:
+                    output[i] = [line]
+                else:
+                    output[i].append(line)
+        #except BadAuthenticationException:     #  <-- got a NameError?
```
codedragon: This will add the key and value (as a list) if not there, and then append to the list, so you can substitute this line for your if statement:

output.setdefault(i, []).append(line)

codedragon:
Looks good.

PR #127, opened 2015-12-10
Updating class project TimeCrunch2 - now with basic functionality! 

JosephVC:
Class project technically does something now.  Some other things as well, as I've not done a pull request in some time.


codedragon:
Having a Readme is always a good idea. An even better idea is to explain in the readme, what the different files are suppose to do, and how to use them. :) I like your idea, but I don't think it does much as is. A good way to start would be to come up with a test, and then write the code to pass the test. I think coming up with tests would help to clarify what you are trying to accomplish.

students/JosephC/IntroProject/TimeCrunch2.py
```diff
@@ -0,0 +1,67 @@
...
+ 
+if squats in your_events: print("You have space to do a set of squats (it'll only take you 30 seconds!).")
+if chair_squats in your_events: print("You have space to a set of squats from your chair (it'll only take you 30 seconds!).")
+if pushups in your_events: print("You have space to do a set of pushups (it'll only take you 45 seconds!).")
+if crunches in your_events: print("You have space to do a set of crunches (it'll only take 14 seconds!).")
+if arm_circles in your_events: print("You have space to a round of arm circles (it'll only take 12 seconds1).")
+if leg_curl in your_events: print("You have space to do a set of leg curls (it'll only take 24 seconds per leg!).")
+
+#Offer the use ability to look through their calendar, converting the wierd datetime format to something more human-readable
+
```
codedragon: Nice find of an applicable library. Does much of what you want to do. However, your interface may leave a little to be desired.

codedragon:
Fyi, generally a good idea to add *.pyc to your .gitignore as they just take up space in a git repository, and can cause silly merge conflicts.

PR #129, opened 2015-12-10
New approach to final project

ryanmorin:
Hey Chris

I've been tinkering with pandas. I can get the program to work... kinda. I'm having trouble with the loop on line 205. The program picks up the last mortgage in the CSV file and calculates it's cash flow. I intended it to loop through all the mortgages. I'm guessing that my indentation is wrong but it could just as easily be something else.

Thanks Chris I enjoyed your class.

### DIFF
--- a/students/ryan_morin/final_project/mortgage.py
+++ b/students/ryan_morin/final_project/mortgage.py
@@ -1,212 +1,50 @@
-__author__ = 'rmorin'
+import datetime as dt
+import pandas as pd
 
-import datetime
-import time
-import calendar
-import math
+class Mortgage(object):
+    """A mortgage from a financial institution. Mortgages have the following properties:
 
-Rates = {'12/25/2014': 1.50, '01/25/2015': 1.55, '02/25/2015': 1.60, '03/25/2015': 1.65, '12/25/2015': 1.75,
-         '12/31/2018': 6.00}
+    Attributes:
+        principal: A float representing the outstanding mortgage amount
+        rate: A float representing the current interest rate
+        dtadv: A string representing the date the mortgage was advanced
+        freq: A string representing the frequency of the mortgage payment
+        am: An int representing the time associated with the mortgage obligation
+    """
+    def __init__(self, principal, rate, dtadv, freq, am, dtmat):
+        """Return a Mortgage object whose principal is *principal*, rate is *rate*, dtadv is *dtadv*,
+        frequency is *freq* and amortization is *am*
+        """
+        self.principal = principal
+        self.rate = rate
+        self.dtadv = dtadv
+        self.freq = freq.upper()
+        self.am = am
+        self.dtmat = dtmat
 
-def Addsemimonth(sourcedate, per):
-    if sourcedate.day == 15 and per % 2 != 0:
-        day = sourcedate.day - 14
-    elif sourcedate.day == 1 and per % 2 != 0:
-        day = sourcedate.day + 14
-    else:
-        day = sourcedate.day
-    month = sourcedate.month + per % 12
-    if month > 12:
-        month = month - 12
-    year = sourcedate.year + (per / 24)
-    return datetime.date(year, month, day)
+    def pmtrate(self):
+        adjrate = self.rate / 100
+        return adjrate
 
-def Addmonth(sourcedate, per):
-    month = sourcedate.month - 1 + per
-    year = sourcedate.year + month / 12
-    month = month % 12 + 1
-    day = min(sourcedate.day, calendar.monthrange(year, month)[1])
-    return datetime.date(year, month, day)
-
-def Addbiweekly(sourcedate, per):
-    dt = sourcedate + (datetime.timedelta(days=14) * per)
-    return datetime.date(dt.year, dt.month, dt.day)
-
-def Addweekly(sourcedate, per):
-    dt = sourcedate + (datetime.timedelta(days=7) * per)
-    return datetime.date(dt.year, dt.month, dt.day)
-
-def rate_date(reqdate, dict):
-    dates = []
-    BootstrpInput = []
-    RevDict = {}
-    posreqdate = 0
-    for key, value in dict.items():
-        TempDate = datetime.date(int(key[6:10]), int(key[0:2]), int(key[3:5]))
-        TempDateNum = str([time.mktime(TempDate.timetuple())])
-        RevDict[TempDateNum] = value
-        dates.append(TempDateNum)
-    ConvReqDate = reqdate
-    DateNum = str([time.mktime(ConvReqDate.timetuple())])
-    dates.append(DateNum)
-    dates.sort()
-    for i in [i for i, x in enumerate(dates) if x == DateNum]:
-        posreqdate = i
-    lower = dates[posreqdate - 1]
-    upper = dates[posreqdate + 1]
-    InvertDates = {value: key for key, value in Rates.items()}
-    BootstrpInput = [InvertDates[RevDict[lower]], RevDict[lower], InvertDates[RevDict[upper]], RevDict[upper]]
-    return BootstrpInput
-
-def DiscRate(RateBookEnds, TargetDate):
-    today = datetime.date.today()
-
-    #Slicing the input table into component parts and assigning to a variable
-
-    LowDateStr = RateBookEnds[0]
-    LowRate = RateBookEnds[1]
-    HighDateStr = RateBookEnds[2]
-    HighRate = RateBookEnds[3]
-
-    #Converting the strings into a datetime format
-
-    LowDate = datetime.date(int(LowDateStr[6:10]), int(LowDateStr[0:2]), int(LowDateStr[3:5]))
-    HighDate = datetime.date(int(HighDateStr[6:10]), int(HighDateStr[0:2]), int(HighDateStr[3:5]))
-    IncDist = abs((TargetDate - LowDate).days)
-    if IncDist == 0:
-        IncDist == 1
-
-    #Calulating the DiscountRate from the BootStrapRate and the YearDecimal
-
-    BootStrapRate = LowRate + (((HighRate - LowRate)/(abs((HighDate - LowDate).days)-1)) * IncDist)
-    YearDecimal = abs((TargetDate - today).days)/365.25
-    DiscountRate = ((1 + BootStrapRate / 200) ** 2) ** (YearDecimal * -1)
-
-    #Returning the DiscountRate for the respective TargetDateStr
-
-    return DiscountRate
-
-def TotalMtgCF(Am, PayFreq, Amt, Rate, MatDateStr, NxtPmtDtStr):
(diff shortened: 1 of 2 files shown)


students/ryan_morin/final_project/mortgage.py
```diff
@@ -1,212 +1,50 @@
...
+import datetime as dt
+import pandas as pd
+
+class Mortgage(object):
+    """A mortgage from a financial institution. Mortgages have the following properties:
+
+    Attributes:
+        principal: A float representing the outstanding mortgage amount
+        rate: A float representing the current interest rate
+        dtadv: A string representing the date the mortgage was advanced
```
PythonCHB: might be good to store this as a datetime object

students/ryan_morin/final_project/mortgage.py
```diff
@@ -1,212 +1,50 @@
...
+        frequency is *freq* and amortization is *am*
+        """
+        self.principal = principal
+        self.rate = rate
+        self.dtadv = dtadv
+        self.freq = freq.upper()
+        self.am = am
+        self.dtmat = dtmat
+
+    def pmtrate(self):
```
PythonCHB: this might be good as a property

students/ryan_morin/final_project/mortgage.py
```diff
@@ -1,212 +1,50 @@
...
+        return adjrate
+
+    def pmtfreq(self):
+        if self.freq == 'MONTHLY':
+            freq = 12
+        elif self.freq == 'BIWEEKLY':
+            freq = 26
+        elif self.freq == 'WEEKLY':
+            freq = 52
         else:
```
PythonCHB: a good idea to set a default freq -- then raise an exception here if it's not one you expect -- not good for someone to have a typeo: weakly -- and then get monthly without any warning..

students/ryan_morin/final_project/mortgage.py
```diff
@@ -1,212 +1,50 @@
...
-            LoopIter = LoopIter + 1
-    elif PayFreq == 'weekly':
-        while NxtPmt < MatDate:
-            NxtPmt = NxtPmt + datetime.timedelta(days=7)
-            LoopIter = LoopIter + 1
+            freq = 12
+        return freq
 
-# Creating a variable called NumPer that holds that number of compounding periods
+    def principal(self):
```
PythonCHB: no need for a method here, the user can jsut do: 

my_mortgage.principle and get it.

also, when you set self.principle in the __init__, you are actually overwriting this method anyway :-)

students/ryan_morin/final_project/mortgage.py
```diff
@@ -1,212 +1,50 @@
...
-            discCF = (etp + eti + ppyamt) * DiscRate(rate_date(Addsemimonth(NxtPmt, item), Rates),
-                                                     Addsemimonth(NxtPmt, item))
-        elif PayFreq == 'biweekly':
-            discCF = (etp + eti + ppyamt) * DiscRate(rate_date(Addbiweekly(NxtPmt, item), Rates),
-                                                     Addbiweekly(NxtPmt, item))
-        else:
-            discCF = (etp + eti + ppyamt) * DiscRate(rate_date(Addweekly(NxtPmt, item), Rates),
-                                                     Addweekly(NxtPmt, item))
-        return TotalMtgCF(Am, PayFreq, Amt, Rate, MatDateStr, NxtPmtDtStr)
+    def amort(self):
```
PythonCHB: same here -- no need for a method simply to return an attribute.

students/ryan_morin/final_project/mtgcalc.py
```diff
@@ -0,0 +1,232 @@
...
+import Mortgage as mc
+import time as t
+
+
+"""
+VARIABLES
+===================
+"""
+
+prepayment_rate = 0.05
```
PythonCHB: if these are essentially constants, it's the convention to give them ALL CAPS names.

students/ryan_morin/final_project/mtgcalc.py
```diff
@@ -0,0 +1,232 @@
...
+default_rate = 0.015
+total = 0
+r0 = 0.04
```
PythonCHB: maybe this set should be in a tuple?

r = (0.04, 0.09, 0.08, ...

then you get r[0], r[1], ...

students/ryan_morin/final_project/mtgcalc.py
```diff
@@ -0,0 +1,232 @@
...
+mtgportfolio = np.array(pd.read_csv('C:/Users/ryan/Desktop/mtgs.csv', index_col=None, header=None, sep=','))
+ycurve = np.array(pd.read_csv('C:/Users/ryan/Desktop/rates.csv', index_col=None, header=None, sep=','))
+
+"""
+COPY OF DATA FRAME
+===================
+"""
+
+ycurve_up = ycurve.copy()
+ycurve_up[2][0] = float(ycurve[2][0]) + r0
```
PythonCHB: here is wehre you can make great use of numpy arrays:

they are native 2-d so you index them as:
ycurve_up[2, 0] -- more efficient.

but you can probably do all this as a "vecgorized operation, somethign like:

ycurve_up[2,:] += r

(if r is an array also, which is should be.)

students/ryan_morin/final_project/mtgcalc.py
```diff
@@ -0,0 +1,232 @@
...
+ycurve_up[2][1] = float(ycurve[2][1]) + r1
+ycurve_up[2][2] = float(ycurve[2][2]) + r2
+ycurve_up[2][3] = float(ycurve[2][3]) + r3
+ycurve_up[2][4] = float(ycurve[2][4]) + r4
+ycurve_up[2][5] = float(ycurve[2][5]) + r5
+ycurve_up[2][6] = float(ycurve[2][6]) + r6
+ycurve_up[2][7] = float(ycurve[2][7]) + r7
+
+ycurve_dn = ycurve.copy()
+ycurve_dn[2][0] = float(ycurve[2][0]) - r0
```
PythonCHB: similarly:
ycurve_dn[2,:] -= r

students/ryan_morin/final_project/mtgcalc.py
```diff
@@ -0,0 +1,232 @@
...
+    return int(term / np.timedelta64(1, 'D'))
+
+
+def num_pmt_made():
+    # Number of payments made
+    pmade = (daysout() / 365) * test.pmtfreq()
+    return int(pmade / np.timedelta64(1, 'D'))
+
+
+def num_pmt_rem():
```
PythonCHB: couldn't these be methods on your Mortgage class?

students/ryan_morin/final_project/mtgcalc.py
```diff
@@ -0,0 +1,232 @@
...
+    rem = num_tot_pmt() - num_pmt_made()
+    return rem
+
+
+def fut_pmt_date(num):
+    if test.pmtfreq() == 52:
+        pmtdate = test.dateadv() + pd.DateOffset(days=(7*num))
+    elif test.pmtfreq() == 26:
+        pmtdate = test.dateadv() + pd.DateOffset(days=(14*num))
+    else:
```
PythonCHB: it's a good idea not to do a blank else -- what if someone passes in something totally weird -- better to check for what you expect, and raise an exception if you get something else.

students/ryan_morin/final_project/mtgcalc.py
```diff
@@ -0,0 +1,232 @@
...
+    elif test.pmtfreq() == 26:
+        freq = '2W'
+    else:
+        freq = 'MS'
+    return freq
+
+
+def mtg_bal_calc():
+    temp = list()
+    for futpmt in range(num_pmt_rem()):
```
PythonCHB: I'm a bit lost on the math here, but I"ll bet you could do this with a numpy operation, rather than a loop.

students/ryan_morin/final_project/mtgcalc.py
```diff
@@ -0,0 +1,232 @@
...
+    The function bootstraps the interest rates between two dates. It then calculates a discount factor (DF). The DF
+    will be used to discount the mortgage cash flows.
+
+    Output
+    ==============
+    Interest rates between two points
+    Time weight (Date - today)/ 365.25
+    Discount rate = (1 + (TmWght /200) ** 0.5) ** -Rates
+    """
+    dates = pd.date_range(ycurve[0][end-1], periods=(int(ycurve[1][end]) - int(ycurve[1][end-1])), freq = 'D')
```
PythonCHB: all these calls to int() and float() make me think you could set the datatypes up the way you want in the first place. For numpy arrays, you can do:

arr = np.array(some_sequence, dtype=np.int32)
or arr = arr.astype(np.int32)

then you're all set from there.
(also dtype = float64, for floats)

students/ryan_morin/final_project/mtgcalc.py
```diff
@@ -0,0 +1,232 @@
...
+    df_dn['TmWght'] = (tmwght / 365.25)
+    df_dn['DiscFctr'] = (1 + (df_dn['TmWght'] / 200) ** 0.5) ** (df_dn['Rates'] * -1)
+    return df_dn
+
+
+"""
+PROGRAM
+===================
+"""
+
```
PythonCHB: this should probably all be in a "if __name__ == "__main__" block. that way you could inport this moodule and use the functions for tests, etc, without running this code.

students/ryan_morin/final_project/mtgcalc.py
```diff
@@ -0,0 +1,232 @@
...
+for end in range(1, 8):
+    yield_c = yield_c.append(yield_curve(end))
+    yield_c_up = yield_c_up.append(yield_curve_up(end))
+    yield_c_dn = yield_c_dn.append(yield_curve_dn(end))
+
+for mtg in range(1, len(mtgportfolio)):
```
PythonCHB: honestly, I'm a bit lost here, but could you do:
for mortgage in mtgportfolio:
    ....
the do the rest with mortgage, rather than continuing to index into mtgportfolio?

students/ryan_morin/final_project/mtgcalc.py
```diff
@@ -0,0 +1,232 @@
...
+    test = mc.Mortgage(float(mtgportfolio[mtg][0]), float(mtgportfolio[mtg][1]), mtgportfolio[mtg][2],
```
PythonCHB: "test" is maybe not the best name here, but this should get you what you want.

Have you printed test out to make sure that it's different each time?

PythonCHB:
Nice work -- looks really useful.

But where are the tests? tsk tsk.

my biggest suggestion is to look into numpy's capabilities -- I"ll bet you could reduce the amount of code here substantially. Using numpy's operations gets you better performance, but that's the least of it -- what it really buys you is more compact code with fewer mistakes.

Ejoy the rest of the program.

-CHB

PR #130, opened 2015-12-10
Final project completed

brendanfogarty27:
Final project makes an API call on a government agency website to grab mortgage application data by state. The program performs a series of data wrangling to make the information more useable, aggregates it in useful ways, and runs two reports, ultimately storing them in a SQLite database. 

The reports try to identify the census tracts that show the strongest evidence of risky lending to low-income borrowers and the most likely areas of discriminatory lending (or red-lining) respectively. 

This project is a easy design for the type of project a local non-profit or government agency might design in researching housing demographics in an area.


students/b_fogarty/Homework/Final Project/project_module_final.py
```diff
@@ -0,0 +1,163 @@
...
+
+        #denial flags (multi-column data function)
+        def denial_flag(row):
+            if row['simplified_decision'] == 'Approved':
+                return 0
+            if row['simplified_decision'] == 'Denied':
+                return 1
+            if row['simplified_decision'] == 'Other':
+                return np.nan
+
```
codedragon: Why do you have all of these closures here? It looks like they could be static methods. This probably won't be a problem here, but closures can have interesting effects on scope, and can cause memory leaks if not configured correctly, so probably best to use them only where necessary.

codedragon:
I don't really understand your testing, and didn't notice anything that might be considered unittesting. I strongly recommend unittesting. It makes it easier to refactor without breaking code, makes it easier for others to understand what your code is suppose to do, and it encourages smaller, more manageable code. Comments are also very helpful. 

I did notice that at least one of your closures from project_module_final.py were re-written in testing, so that would be another reason to just have them as regular functions, so they can be re-used. 

I noticed in one place you asked for input in all caps, but it is just as easy to convert to all caps using string.upper().

It is helpful in your readme, to say things like, start by running file x, inputting x, and this will give you y, which you can then use for z. Iow, a how-to for your code. This other code is for support for x, to run tests do this, etc. It also helps if you don't commit lots of extraneous files, like *.pyc and probably the sqlite files. These can be filtered in .gitignore. 

Overall, looks like useful code, and I hope it deters predatory loans.:)

PR #131, opened 2015-12-11
Class project for review

gabrielx52:
Its far from polished, but I feel like that would take me much longer than the time available.

Its a inventory/par setter with a 'pack list' generator. 
The user inputs items and sets pars for them and Its saved to a json file in the current dir. It also generates a pack list depending on the current inventory of the items. There are a few more features I'd like to eventually add including using the measurements package.. but that is out of my league for now.

I had an issue trying to write tests for it. I started out with writing tests before the code but I wasn't sure how to tackle the menus/ how to simulate user input... I've included one test but to get it to work a line needs to be hashed out in the main code. Is there is standard method for writing tests for something like this?

Gabriel


PythonCHB:
Testing interactive programs is a nightmare. A couple options are:

1) use the pexpect pacakge: 
https://pexpect.readthedocs.org/en/stable/

that lets you automate (and test) command line interaction

2) override your functions that ask for user ineteraction, so you can teset the rest.

3) don't actually test the interaction (at least not automatically. Rather, make sure the code that asks for user interaction is isolated, so you can test all the logic code without any interaction. That's what I usually end up doing.

students/gabriel_meringolo/project/pack_list2.py
```diff
@@ -0,0 +1,167 @@
+import json
+from os.path import dirname, realpath, join
+from os import listdir
+from collections import OrderedDict
+from fractions import Fraction
```
PythonCHB: nice use of some of the less obvious stuff in the standard library.

students/gabriel_meringolo/project/pack_list2.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def main_menu():
+    """
+    main menu
+    """
+    menu_choice = input("\nPack List V1.0\n---------------\n"
+                        "1. Use Existing List\n"
+                        "2. Create New List\n"
+                        "3. Exit\n\n>>> ")
```
PythonCHB: I usually put a .strip() on here, so a inadvertent space won't mess up the input.

students/gabriel_meringolo/project/pack_list2.py
```diff
@@ -0,0 +1,167 @@
...
+    if menu_choice == "1":
+        use_existing()
+    elif menu_choice == "2":
+        create_new_list(input("\nEnter new list name: "))
+    elif menu_choice == "3":
+        exit_pl()
+    else:
+        print("Not a valid selection, please enter 1, 2, or 3\n\n ")
+        main_menu()
```
PythonCHB: this is doing a recursive call -- it may cause problems if run enough -- it's usually best to use a "while True": loop to keep the input going. See my solution to the Mailman exercise.

students/gabriel_meringolo/project/pack_list2.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def current_lists():
+    """
+    displays all .json files in current dir
+    """
+    print("\nCurrent Lists:\n--------------")
+    for files in listdir(location):
```
PythonCHB: careful with variable names -- if it's one file, use a singular name : 'filename', maybe. "file" is a built in so better not to use that.

students/gabriel_meringolo/project/pack_list2.py
```diff
@@ -0,0 +1,167 @@
...
+        sub_menu(list_name)
+    else:
+        make_choice = input("\n\aList does not exist, create new list? Y/N ")
+        if make_choice.lower() == "y":
+            create_new_list(list_name)
+        elif make_choice.lower() == "n":
+            main_menu()
+        else:
+            print("Not a valid selection, returning to main menu.")
+            main_menu()
```
PythonCHB: again, this is recursive and could blow up if running along time. better to return from this function when done.

students/gabriel_meringolo/project/pack_list2.py
```diff
@@ -0,0 +1,167 @@
...
+                       "4. Return To Main Menu\n"
+                       "5. Exit Program\n\n>>> ")
+    if sub_choice == "1":
+        view_list(list_name)
+    elif sub_choice == "2":
+        edit_list(list_name)
+    elif sub_choice == "3":
+        pack_list(list_name)
+    elif sub_choice == "4":
+        main_menu()
```
PythonCHB: same here -- each call to a menu-processing function should simply return -- maybe with a flag as to what it did, but better not to call main_menu from inside function. -- the stack is going to just keep growing.

students/gabriel_meringolo/project/pack_list2.py
```diff
@@ -0,0 +1,167 @@
...
+    input("\nHit 'Enter' to continue.\n")
+    sub_menu(list_name)
+
+
+def create_new_list(list_name):
+    """
+    creates new .json file containing '{}'
+    """
+    with open(join(location, "{}.json".format(list_name)), "w") as f:
+        json.dump({}, f)
```
PythonCHB: why create a json file with an empty dict in it?

to make all this more future proof, you may want a class for a list. then you can create an empty one, and save it to json, and load it from json, but also load/save other ways in the future. and there may be some logic to attach to it as well.

students/gabriel_meringolo/project/pack_list2.py
```diff
@@ -0,0 +1,167 @@
...
+def pack_list(list_name):
+    """
+    generates pack list
+    """
+    data = read_json(list_name)
+    par_check = OrderedDict()
+    print("\n")
+    for i in data:
+        current_amt = input("Current amount of {}: ".format(i))
+        if "/" in current_amt:
```
PythonCHB: this is pretty slick!

students/gabriel_meringolo/project/pack_list2.py
```diff
@@ -0,0 +1,167 @@
...
+
+def edit_list(list_name):
+    """
+    edits inventory list
+    """
+    while True:
+        item = input("Enter new item: ")
+        par = input("Set par: ")
+        if "/" in par:
+            par = str(round(float(Fraction(par)),2))
```
PythonCHB: this is a one-liner, bbut may be worth putting in a function none the less.

students/gabriel_meringolo/project/pack_list2.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    updates json file
+    """
+    data = read_json(list_name)
+    data[item] = par
+    with open(join(location, "{}.json".format(list_name)), "w") as f:
+        json.dump(data, f, indent=4)
+
+
+def read_json(list_name):
```
PythonCHB: these last two you should be able to test, at least...

students/gabriel_meringolo/project/test_pack_list2.py
```diff
@@ -0,0 +1,9 @@
+import pack_list2
```
PythonCHB: if your function that get input from the user were structured to return the input rather than call another function, then it would be much easeir to test stuff. For isntnce, you could overwrite submenu in your test file:

old_submenu = pack_list2.submenu
def submenu():
    return 2
pack_list2.submenu = sumenu
[test code that uses submenu]
pack_list2.submenu = old_submenu

Also -- your functions that display something could return the text that is going to be displayed, e=rather than printing directly, so they could be tested. then you would call them with:

print(view_list)

students/gabriel_meringolo/session_09/generator.py
```diff
@@ -0,0 +1,36 @@
...
+        yield next
+        start, next = next, start + next
+    return None
+
+
+def prime():
+    start = 2
+    while start < 40:
+        yield start
+        pass
```
PythonCHB: three passes in a row ? ;-)

PythonCHB:
Nice work -- you're getting the hang of it for sure.

Enjoy the rest of the program.

-Chris

PR #132, opened 2015-12-11
Final Project: Culture Filter

robalford:
This program will be more useful once I add more scrapers, but there are enough here to get a sense of how it will work. For my class project, I wanted to focus on the design of the program and would love to get some feedback on the OOP aspects in particular, as I am still figuring a lot of this stuff out. See my notes and questions in the comments. Thanks!


students/robalford/culture_filter/culture_filter.py
```diff
@@ -1,173 +1,237 @@
...
+        filename: The name of the file used for storage as a string.
+        recommendations: A data strucutre to store all recommended works
+        for this medium.
+    """
+
+    # Abstract class attributes for inheritance to various mediums.
+    # Is this the correct way to handle this situation in Python?
+    medium = ''
+    filename = ''
+
```
codedragon: It would be better to have these in init as self.medium and self.filename, so that your subclasses don't share these attributes. Probably doesn't matter here, as they will probably be overwritten, but better practice in general.

students/robalford/culture_filter/culture_filter.py
```diff
@@ -1,173 +1,237 @@
...
+    #         "reviewed by": ['Paste Music']
+    #     },
+    # }
 
     def __init__(self, recommendations=None):
+        """Load recommendation data into recommendations attribute."""
         try:
             self.recommendations = self.load_data()
         except FileNotFoundError:
             self.recommendations = {}
```
codedragon: When you catch the fileNotFoundError, you should probably let the user know that there was an issue with a file, and the database is empty.

students/robalford/culture_filter/culture_filter.py
codedragon: I am confused by your use of the variable medium. It is initialized in lots of different places, sometimes as a string, sometimes initializing a class object. It appears to not ever be used anywhere as a string, so I think you can get rid of it in the class Recommender entirely.

students/robalford/culture_filter/culture_filter.py
```diff
@@ -1,173 +1,237 @@
...
+        recommendation has already been recorded, but was gathered by the
+        scraper again on this pass, move on to the next recommendation."""
+        # Call __init__ here to load current recommendations. Is there a better
+        # way to handle this?
+        self.medium.__init__()
         for r in self.recommendations:
             work = '{}: {}'.format(r[0].lower(), r[1].lower())
-            if work in self.medium.recommendations:
+            if work not in self.medium.recommendations:
+                self.medium.recommendations[work] = {
```
codedragon: calling medium.__init__() here is not doing anything, you have already initialized. I think the problem here is that the super class (Publication) should not know anything about the subclasses. You are collecting your objects in all_publications, and when you iterate through it calling add_recommendations, you could send in the object's dictionary, so that self.medium.recommendations is just an input variable. I'm sure there are other ways to deal with this, but the thing you want to accomplish is not referencing a subclass in a superclass.

students/robalford/culture_filter/culture_filter.py
```diff
@@ -1,173 +1,237 @@
...
+                scrape(*args, **kwargs)
+            # Will eventually narrow in on specific exceptions, but for
+            # now, it's a catch-all to make sure the program keeps running
+            # if one of the scrapers fails
+            except Exception as e:
+                print("""There was a problem with one of your scrapers.
+                    See error below:""")
+                print(e)
+                return
+        return exception_catcher
```
codedragon: I think if you use the catch_scraper_exception decorator with an input variable, it will not have to be static. This may be helpful: http://www.artima.com/weblogs/viewpost.jsp?thread=240845

codedragon:
I hope my comments are helpful. Happy to see some tests. Makes it easier to refactor and make sure things still work. Looks useful. I like how you have separated the file interface, this way you could switch to a database or network calls if you needed to.

PR #133, opened 2015-12-12
project file

HasMikeW:
doesn't quite work yet.


PythonCHB:
well, working code is good code -- but almost working code is a good start!

students/mike_waddle/project/yahtzee.py
```diff
@@ -0,0 +1,153 @@
...
+    dice =[random.randint(1,6) for x in range(numDice)]
+    return dice
+
+
+
+def scorecard(dice):
+    category = input("Enter the category you want to input your score: ")
+    dicecount = Counter(dice)
+    keylist = list(dicecount.keys())
+    valuelist = list(dicecount.values())
```
PythonCHB: why do you need these as lists?

students/mike_waddle/project/yahtzee.py
```diff
@@ -0,0 +1,153 @@
...
+    house = 0
+    smstr = 0
+    lgstr = 0
+    yahtzee = 0
+    chance = 0
+    bottotal = trips + quads + house + smstr + lgstr + yahtzee
+    total = toptotal + bottotal
+    score = 0
+    
+    if category == 1:
```
PythonCHB: this sure looks like logic that could be put in one line, rather than a pile of if statements. maybe if you put it all in a cdict, rather than variables:

{'ones = 0,
 'twos = 0,
 'threes = 0,
 'fours': 0,}

or better yet, keep the as numerical:

{1:0,
2:0,
3:0,
...
}
and convert to text for display.

then you can do math on it a whole lot easier.

students/mike_waddle/project/yahtzee.py
```diff
@@ -0,0 +1,153 @@
...
+    elif category == 6:
+        sixes = dice.count(6) * 6
+
+    elif category == 7 and 3 in valuelist:
+        trips = sum(dice)
+
+    elif category == 8 and 4 in valuelist:
+        quads = sum(dice)
+
+    elif category == 9 and (3 and 2 in valuelist):
```
PythonCHB: so now it gets more complicated, but you might as well do the easy ones the easy way.

students/mike_waddle/project/yahtzee.py
```diff
@@ -0,0 +1,153 @@
...
+    elif category == 11 and len(keylist) > 4:
+        lgstr = 40
+
+    elif category == 12 and len(keylist) == 5:
+        yahtzee = 50 
+
+    elif category == 13:
+        chance = sum(dice)
+
+    print(card)
```
PythonCHB: probably a good idea to have all the display in a separate function. - then you can test, etc...

students/mike_waddle/project/yahtzee.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+
+def main():
+    
+    roll1 = roll(5)
+
+    print(roll1)
+
+    keep1 = [int(x) for x in input("Enter the dice you want to keep 1-5 and separate by comma or 9 to score or 0 to quit: ").split(',')]
```
PythonCHB: kinda easy to mess up the input -- so maybe put this in a try: blcok, so you can at least give a nice error msg.

PythonCHB:
you've got a nice start here. I'd try to think about how you can generalize the code a bit more. rather than simply hard-coding each possibility. Maybe use functions to compute each type of scoring -- then you could run an roll through a list of functions....

Enjoy the rest of the program!

PR #134, opened 2015-12-12
Project 

elaewin:
Well, it's much shorter than I wanted it to be. I was unable to get a lot of the elements to work. the todo.txt file has a list of things I was hoping to achieve, but wasn't able to manage.  The 'final' minigame is in the abridged folder.


PythonCHB:
Careful with committing files that you shouldn't -- you really nicer want to do a plain "git add", alwayas specifically specify what you want to add: git add *.py, and the like.
and/or use a good .gitignore -- there are a bunch here:

https://github.com/github/gitignore

if you merged the python one with the OS-X one, it would make this easier.

students/erica_winberry/project/bbjungle/abridged/npcs/teachers.py
```diff
@@ -0,0 +1,72 @@
+# coding=utf-8
+"""
+Creatures living in the central town.
+
+'Tale' mud driver, mudlib and interactive fiction framework
+Copyright by Irmen de Jong (irmen@razorvine.net)
+"""
+from __future__ import absolute_import, print_function, division, unicode_literals
```
PythonCHB: are you using python2 here? in which case, this is a good way to get py3 compatible code.

students/erica_winberry/project/bbjungle/abridged/npcs/town_creatures.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class WalkingRat(NPC):
+    def init(self):
+        super(WalkingRat, self).init()
+        mud_context.driver.defer(2, self.do_idle_action)
+        mud_context.driver.defer(4, self.do_random_move)
+        self.aggressive = True
+
+    def do_idle_action(self, ctx):
```
PythonCHB: it might be worth it to generalize this -- you could have a IdleAction class, and it would be callable. you could pass in the variable idle actions and it would figure out the randomization based on the number of actions, etc.

A bit less repeated code....

students/erica_winberry/project/bbjungle/abridged/npcs/town_creatures.py
```diff
@@ -0,0 +1,102 @@
...
+        if random.random() < 0.5:
+            self.tell_others("{Title} wiggles %s tail." % self.possessive)
+        else:
+            self.tell_others("{Title} sniffs around and moves %s whiskers." % self.possessive)
+        ctx.driver.defer(random.randint(5, 15), self.do_idle_action)
+
+    def do_random_move(self, ctx):
```
PythonCHB: this doesn't look very specific to this particular class -- put it in a base class?

students/erica_winberry/project/bbjungle/abridged/npcs/town_creatures.py
```diff
@@ -0,0 +1,102 @@
...
+        # This is the preferred way (it's efficient).
+        mud_context.driver.defer(2, self.do_cry)
+
+    def do_cry(self, ctx):
+        self.tell_others("{Title} mutters to herself as she putters around behind the library counter.")
+        ctx.driver.defer(random.randint(20, 40), self.do_cry)
+
+    def notify_action(self, parsed, actor):
+        greet = False
+        if parsed.verb in ("hi", "hello"):
```
PythonCHB: this also looks like pretty common code -- put in a subclass? or a mixin?

students/erica_winberry/project/bbjungle/abridged/play bbj (gui).pyw
```diff
@@ -0,0 +1,27 @@
...
+import os
+
+try:
+    import tale.driver
+    tale_from_lib = True
+except ImportError:
+    tale_from_lib = False
+
+if not tale_from_lib:
+    if os.path.exists("../../tale/__init__.py"):
```
PythonCHB: this kind of os.path  manipulation is pretty fragile. It's really best to make a package of your code, and install it -- in develop mode while you are working on it.

students/erica_winberry/project/bbjungle/testing/play bbj (gui).pyw
```diff
@@ -0,0 +1,27 @@
...
+
+try:
+    import tale.driver
+    tale_from_lib = True
+except ImportError:
+    tale_from_lib = False
+
+if not tale_from_lib:
+    if os.path.exists("../../tale/__init__.py"):
+        sys.path.insert(0, os.path.abspath("../.."))
```
PythonCHB: again, getting ugly. We didn't cover packaging on class, but this calls for it. This is a good place to start:

https://python-packaging-user-guide.readthedocs.org/en/latest/

PythonCHB:
Nice work -- this looks pretty cool!

I'm not familiar with the game framework, so not much to say, but this looks like pretty clean code.

Enjoy the rest of the program!

